PR #3, opened 2023-06-08
Cull idle kernels

ScottDelbecq:
With the actual settings of the server deployment, idle kernels were kept until the user closes the tab. This PR sets a limit of 20 mins before culling the idle kernels.


thomasplanes (changes requested):
I think 2 hours before deleting the user session would be great

ScottDelbecq:
Thanks lets try 2 hours first then!

thomasplanes (approved):

PR #4, opened 2023-07-07
Update fleet model

ScottDelbecq:
This PR aims at integrating updates for the fleet model:

- The computation now starts from an empty dataframe to avoid bugs when running several computations
- The freight model is now common to the simple or complex fleet efficiency models


thomasplanes (approved):
This PR includes all the required commits for improving fleet renewal and freight efficiency models .

PR #10, opened 2023-08-23
Add temperature increase due to CO2 and non-CO2 effects

thomasplanes:
This PR allows decompising temperature increase due to CO2 and non-CO2 effects. A correction of the corresponding figure has been added.


ScottDelbecq (changes requested):
Thank you for this PR.
As mentioned, the modification of the notebooks are not related to this PR and must thus be cancelled.

PR #12, opened 2023-09-01
Correction of multiple run issue

### LINKED ISSUE #11: Multiple computes only provides outputs of disciplines affected by inputs change
GEMSEO recomputes only disciplines whose inputs have changed between computations. AeroMAPS starts from an empty Dataframe at each computation and concatenate the Dataframe of each discipline. As a result, the Dataframe of uncomputed disciplines are empty and their outputs are missing in the next computation.

ScottDelbecq:
This PR resolves #11 by stopping from an empty Dataframe at each run. Also, some refactoring was made for more clarity and efficiency in data handling.


thomasplanes (approved):
It allows the execution of multiple run by correcting previous bug.

PR #53, opened 2023-12-19
Improve climate model based on GWP*

thomasplanes:
- Improve climate model based on GWP*
- Add a detailed Excel file


antSGS (approved):
Climate models PR approved, no issue found in testing notebooks

ScottDelbecq (approved):

PR #67, opened 2024-03-28
Custom input file

ScottDelbecq:
This PR enables to use custom input files by defining the path in the configuration file. An example of usage is added to the main notebook.


thomasplanes (approved):
It works, an issue will be added for providing dedicated notebooks

PR #81, opened 2024-05-29
Feature added: simplified energy models

antSGS:
This new feature enables the user to use much more simple energy cost models.

These are based on simple MFSP interpolations. They are calibrated using default bottom-up (detailed) energy cost model results. 
However, as with other models, any user can indicate its own assumptions. 

However, note that these models are not able to compute investment chronologies, detailled MFSP breakdowns or abatement costs.


antSGS:
No idea why the checks are not sucessfull

PR #90, opened 2025-02-19
Life cycle assessment

felixpollet:
LCA capabilities for AeroMAPS.


antSGS:
Notebooks tested locally, but LCA notebooks not running as depending on optional install and private credentials. 
PR approved but need to modify tests workflow for further PRs.

antSGS (approved):
See my previous comment.

PR #97, opened 2025-04-04
Fixes binder not running issue

antSGS:
Using poetry instead of pip+git
Forces poetry to install in binder "base" env
@ScottDelbecq check if that's seems ok please?

### DIFF
--- a/binder/postBuild
+++ b/binder/postBuild
@@ -1,1 +1,3 @@
-pip install git+https://github.com/AeroMAPS/AeroMAPS.git
+pip install poetry
+poetry config virtualenvs.create false --local
+poetry install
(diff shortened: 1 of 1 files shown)


binder/postBuild
```diff
@@ -1 +1,3 @@
-pip install git+https://github.com/AeroMAPS/AeroMAPS.git
\ No newline at end of file
+pip install poetry
```
ScottDelbecq, line 1: I would fix the poetry version like `pip install poetry==1.8.3` for more stability.

ScottDelbecq (changes requested):
Thanks for this PR! Just a minor change requested.

PR #100, opened 2025-04-09
Corrected poetry dynamic versioning

ScottDelbecq:
This PR (re)-introduces the poetry dynamic versioning mechanism as it is used in the release workflow.


antSGS (approved):
Dynamic versionning OK

PR #114, opened 2025-12-05
Default and custom LCA models + New configurations files

thomasplanes:
- Add default and custom LCA models
- Add new configuration files for a generic use
- Update notebooks
- Prepare additional features for v1


felixpollet (approved):

antSGS:
Computation time monitoring does not run, is it linked with many renaming that were performed?

aeromaps/utils/functions.py
```diff
@@ -233,54 +233,10 @@ def create_partitioning(file, path=""):
...
-        "commercial_aviation_coefficient": commercial_aviation_coefficient_partitioned,
-        "historic_start_year": historic_start_year_partitioned,
-        "prospection_start_year": prospection_start_year_partitioned,
-    }
-    partitioned_inputs_path = pth.join(path, "partitioned_inputs.json")
-    with open(partitioned_inputs_path, "w") as outfile:
-        json.dump(partitioned_inputs_dict, outfile)
-
-    # Create a CSV file for initialisation of vector inputs.
-    # TODO: not necessary without optim, check relevance of the process?
```
antSGS, line 266: Were you able to understand why ? At least from the test it semms to work well so ok for me.

aeromaps/notebooks/publications/joas_2023/data/outputs.json
antSGS, line 1: Check if the offset is normal for some vectors (e.g. temperature increse now starts with a zero)

antSGS (approved):
Thanks for the work, it seems to work fine, despite small modifications in outputs that are most likely related to models change.

thomasplanes:
> Computation time monitoring does not run, is it linked with many renaming that were performed?

The explanation is the renaming of the folders for the tutorial notebooks. It should be OK for the next PR (release of v1.0.0) but we can change this test from a failed test to an information.

aeromaps/notebooks/publications/joas_2023/data/outputs.json
thomasplanes: The difference comes from the type of climate models in AeroCM (e.g. GWP vs FaIR). A dedicated issue has been added in AeroCM for improving consistency between climate models, but this does not call this version into question (the differences being minor anyway).

aeromaps/utils/functions.py
thomasplanes: vector_inputs is still present, which enables optimisation to work

ScottDelbecq (approved):
All good!

PR #134, opened 2026-03-23
Improve error handling across the codebase

### LINKED ISSUE #46: Added a way to find the model involved in the error
It is somethings difficult to identify models that yield an error. It would be nice to have at leat the name of the model.

ScottDelbecq:
This PR adds error handling improvements across 6 files of the project, making failures easier to diagnose and debug. It therefore resolves #46

### DIFF
--- a/aeromaps/core/process.py
+++ b/aeromaps/core/process.py
@@ -1348,8 +1348,9 @@ class AeroMAPSProcess(object):
                             model = AeroMAPSAutoModelWrapper(model=model)
                         self.disciplines.append(model)
                     else:
-                        logging.warning(
-                            "Model '%s' has no compute() method and will be skipped.", model.name
+                        raise AttributeError(
+                            f"Model '{model.name}' has no compute() method. "
+                            "All AeroMAPSModel subclasses must implement compute()."
                         )
                 else:
                     logging.warning(
--- a/aeromaps/core/process.py
+++ b/aeromaps/core/process.py
@@ -1158,12 +1158,6 @@ class AeroMAPSProcess(object):
                     model._initialize_df()
                     if hasattr(model, "pathways_manager") and hasattr(model, "custom_setup"):
                         # TODO harmonise the way to pass the pathways manager with generic models
-                        if not hasattr(self, "pathways_manager"):
-                            raise RuntimeError(
-                                f"Model '{model.name}' requires a pathways_manager but "
-                                "generic energy has not been initialized. "
-                                "Call setup_mda() before _initialize_disciplines()."
-                            )
                         model.pathways_manager = self.pathways_manager
                         model.custom_setup()
                     if hasattr(self, "fleet_model"):
--- a/aeromaps/core/process.py
+++ b/aeromaps/core/process.py
@@ -1438,8 +1438,8 @@ class AeroMAPSProcess(object):
                     for key, value in data.items():
                         if key in merged_data:
                             logging.warning(
-                                "Parameter '%s' defined in multiple input files; "
-                                "the value from '%s' will be used.",
+                                "Parameter '%s' is defined in multiple input JSON files; "
+                                "the value from file '%s' will override the previous one.",
                                 key, json_file,
                             )
                         merged_data[key] = value
@@ -1641,8 +1641,11 @@ class AeroMAPSProcess(object):
                     setattr(self.parameters, field_name, new_value)
                 else:
                     logging.warning(
-                        "Field '%s' has an unexpected size %d and will not be reformatted.",
-                        field_name, field_value.size
+                        "Field '%s' has an unexpected size %d (expected %d or %d); "
+                        "cannot reformat it to a time-indexed Series.",
+                        field_name, field_value.size,
+                        self.parameters.end_year - self.parameters.climate_historic_start_year + 1,
+                        self.parameters.end_year - self.parameters.historic_start_year + 1,
                     )
 
     def _update_variables(self):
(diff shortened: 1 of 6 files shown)


aeromaps/core/process.py
```diff
@@ -1298,22 +1158,37 @@ def check_instance_in_dict(d):
...
-                        if model.model_type == "custom":
+                        if getattr(model, "model_type", "auto") == "custom":
                             model = AeroMAPSCustomModelWrapper(model=model)
                         else:
                             model = AeroMAPSAutoModelWrapper(model=model)
                         self.disciplines.append(model)
                     else:
-                        print(model.name)
+                        logging.warning(
+                            "Model '%s' has no compute() method and will be skipped.", model.name
```
antSGS: It is not normal not to have a compute. I suggest that we throw an error instead?

aeromaps/core/process.py
```diff
@@ -1298,22 +1158,37 @@ def check_instance_in_dict(d):
...
+                        )
                 else:
-                    print(f"{key} is not an instance of AeroMAPSModel")
+                    logging.warning(
+                        "Entry '%s' is not an AeroMAPSModel instance and will be skipped.", key
```
antSGS: Same comment here. A non AeroMAPSModel is likely an error no?

aeromaps/core/process.py
```diff
@@ -1298,22 +1158,37 @@ def check_instance_in_dict(d):
                     model._initialize_df()
                     if hasattr(model, "pathways_manager") and hasattr(model, "custom_setup"):
                         # TODO harmonise the way to pass the pathways manager with generic models
+                        if not hasattr(self, "pathways_manager"):
```
antSGS, line 1329: I do not understand how we could end up here as it is encapsulated in a if hasattr(model, "pathways_manager") and hasattr(model, "custom_setup") ??

aeromaps/core/process.py
```diff
@@ -1380,14 +1255,26 @@ def _initialize_inputs(self, use_defaults=True):
...
+                        )
+                    except json.JSONDecodeError as e:
+                        raise json.JSONDecodeError(
+                            f"Invalid JSON in '{file_path}': {e.msg}", e.doc, e.pos
+                        ) from e
+                    for key, value in data.items():
+                        if key in merged_data:
+                            logging.warning(
+                                "Parameter '%s' defined in multiple input files; "
+                                "the value from '%s' will be used.",
```
antSGS: Question on my side: what will %s display here?

aeromaps/core/process.py
```diff
@@ -1575,7 +1427,10 @@ def _format_input_vectors(self):
                     new_value = pd.Series(field_value, index=index)
                     setattr(self.parameters, field_name, new_value)
                 else:
-                    print(f"Field {field_name} has an unexpected size {field_value.size}")
+                    logging.warning(
+                        "Field '%s' has an unexpected size %d and will not be reformatted.",
```
antSGS: Message clear enough? "Cannot be reformated?"

antSGS (changes requested):
Thanks for the PR @ScottDelbecq. I left a few minor comments and questions.

aeromaps/core/process.py
ScottDelbecq: It logs key (the parameter name) and json_file (the filename).

aeromaps/core/process.py
ScottDelbecq: Corrected in [d24a4f7](https://github.com/AeroMAPS/AeroMAPS/pull/134/commits/d24a4f7a949b1db5174ac368488ec18f34498cc1)

aeromaps/core/process.py
ScottDelbecq: Corrected in [457440a](https://github.com/AeroMAPS/AeroMAPS/pull/134/commits/457440a02a3935d6a00a8f228574d89bf5863cae)

aeromaps/core/process.py
ScottDelbecq: Corrected in [3f8ecbf](https://github.com/AeroMAPS/AeroMAPS/pull/134/commits/3f8ecbf8ec4129b2d9fb20c328dd2467be274931)

aeromaps/core/process.py
ScottDelbecq: You are right! My bad. Corrected in [f1b9154](https://github.com/AeroMAPS/AeroMAPS/pull/134/commits/f1b9154c5efbcdd751eac94725ab36f829b7efbe)

antSGS: My bad, you were perfectly right. We are talking about AeroMAPSProcess pathways_manager attribute, which maybe not properly initialized (though it will likely throw an error somewhere else if setput_mda was not called)

You can cherry pick this commit?

antSGS (changes requested):

aeromaps/core/process.py
antSGS: OK

aeromaps/core/process.py
antSGS: Thanks

aeromaps/core/process.py
antSGS: All right, i'm used to f-strings now i forgot the % synthax...

aeromaps/core/process.py
antSGS: perfect

antSGS (changes requested):
Thnaks, all good

antSGS (approved):
Finaly good, thanks.

PR #139, opened 2026-04-16
Two new models for estimating air traffic demand for passengers

iancostalves:
This PR adds two new models for estimating passenger demand (RPK) from population, GDP per capita, and energy costs:
- One model is based on income and price elasticities, supposed to be constant in time
- Another is based on logistic functions to estimate per-capita RPK as a function of per-capita GDP, which are then corrected by a price elasticity (supposed to be constant in time)

Two application notebooks are also added:
- One for comparing the two models from 2020 until 2100, with 3 different price assumptions
- Another for comparing FOSSIL and SAF scenarios from 2020 until 2070, with 3 different background scenarios based on the Shared Socioeconomic Pathways

Some extra code modifications were required to correctly initialize and the MDA, to include population and gdp as vector inputs for the historical period, to include a few interesting plots, and some minor modifications in figure sizes.


aeromaps/notebooks/publications/wctr_2026/models/socioeconomic_drivers.py
```diff
@@ -0,0 +1,70 @@
...
+        super().__init__(name=name, *args, **kwargs)
+
+    def compute(
+        self,
+        population_reference_years: list,
+        population_reference_years_values: list,
+        gdp_per_capita_reference_years: list,
+        gdp_per_capita_reference_years_values: list,
+        covid_end_year_passenger: int,
+    ) -> Tuple[pd.Series, pd.Series]:
```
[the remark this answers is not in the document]
iancostalves: @copilot apply changes based on [this feedback](https://github.com/AeroMAPS/AeroMAPS/pull/139#discussion_r3094377915)

aeromaps/core/models.py
```diff
@@ -177,6 +181,28 @@
...
+
+models_traffic_constant_elasticities = {
+    "rpk_constant_elasticity": RPKConstantElasticity("rpk_with_elasticity"),
+    "rpk_measures": RPKMeasures("rpk_measures"),
+    "rpk_reference": RPKReference("rpk_reference"),
+    "total_aircraft_distance": TotalAircraftDistance("total_aircraft_distance"),
+    "rtk": RTK("rtk"),
+    "rtk_reference": RTKReference("rtk_reference"),
+    "ask": ASK("ask"),
+}
```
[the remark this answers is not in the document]
iancostalves: @copilot apply changes based on [this feedback](https://github.com/AeroMAPS/AeroMAPS/pull/139#discussion_r3094378011)

aeromaps/plots/single_scenario/costs.py
```diff
@@ -532,6 +532,299 @@ def _update_plot_elements(self):
...
+                "carbon_tax": energy
+                * self.df.loc[y, f"{p.name}_mean_unit_carbon_tax"].fillna(0).values
+                / total_rpk,
+                "tax": energy
+                * self.df.loc[y, f"{p.name}_mean_unit_tax"].fillna(0).values
+                / total_rpk,
+                "subsidy": energy
+                * self.df.loc[y, f"{p.name}_mean_unit_subsidy"].fillna(0).values
+                / total_rpk,
+            }
```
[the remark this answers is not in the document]
iancostalves: @copilot apply changes based on [this feedback](https://github.com/AeroMAPS/AeroMAPS/pull/139#discussion_r3094378036)

aeromaps/plots/single_scenario/costs.py
```diff
@@ -532,6 +532,299 @@ def _update_plot_elements(self):
...
+        self.ax.set_title("Energy costs per RPK – Breakdown by carrier")
+        self.ax.set_xlabel("Year")
+        self.ax.set_ylabel("Energy costs [€/RPK]")
+        self.ax.set_xlim(self.prospective_years[0], self.prospective_years[-1])
+
+    def _update_plot_elements(self):
+        for collection in self.ax.collections:
+            collection.remove()
+        for line in self.ax.lines:
+            line.remove()
```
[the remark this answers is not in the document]
iancostalves: @copilot apply changes based on [this feedback](https://github.com/AeroMAPS/AeroMAPS/pull/139#discussion_r3094378070)

aeromaps/core/gemseo.py
```diff
@@ -169,6 +169,11 @@ def update_defaults(self):
             #     self.default_inputs[input] = array([0])
             if hasattr(self.model.parameters, input):
                 self.default_input_data[input] = getattr(self.model.parameters, input)
+        # Also register coupling defaults from the model (seed values for MDA initialization)
+        if hasattr(self.model, '_coupling_defaults'):
```
antSGS, line 213: Nice feature ! How does its work though ? You have to specify these as model attributes?

aeromaps/core/process.py
```diff
@@ -1614,6 +1625,8 @@ def _format_input_vectors(self):
                 "freight_init",
                 "energy_consumption_init",
                 "total_aircraft_distance_init",
+                "gdp_per_capita_init",
```
antSGS, line 1692: So these become mandatory inputs now ?

aeromaps/core/process.py
```diff
@@ -634,7 +634,8 @@ def list_str_inputs(self):
         """
         return self.data["str_inputs"]
 
-    def plot(self, name, save=False, size_inches=None, remove_title=False):
+    def plot(self, name, save=False, size_inches=None, remove_title=False,
```
antSGS, line 660: Remains from the last PR haha?

aeromaps/models/air_transport/air_traffic/price_and_income_elasticity.py
```diff
@@ -0,0 +1,311 @@
...
+        self.sigma: float = 0.0015018367707702585
+        self.income_elast: float = 1.318374753418146
+        self.price_elast: float = -0.2808214954796415
+        # Exchange rate used to convert doc_all_energy_costs_per_rpk from EUR to USD [EUR/USD]
+        self.eur_usd_exchange_rate: float = 0.9
+
+    def _initialize_df(self):
+        super()._initialize_df()
+        # Seed value for MDA coupling initialization: approximate 2019 all-energy cost per RPK
+        self._coupling_defaults = {
```
antSGS, line 40: Same comment as in gemseo.py, i'm eager to understand how it works. So far i think couplings defaults were passed in as vector inputs, it means your PR suggests a new standard way of doing that ?

aeromaps/models/air_transport/air_traffic/rpk_constant_elasticity.py
```diff
@@ -0,0 +1,311 @@
+from typing import Tuple
+
+import pandas as pd
+
+from aeromaps.models.base import AeroMAPSModel
+
+
+class RPKConstantElasticity(AeroMAPSModel):
```
antSGS: I think we need to clarify the model and file names with the other price_elasticity.py model

aeromaps/models/impacts/costs/airlines/direct_operating_costs.py
```diff
@@ -2264,6 +2285,22 @@ def compute(
         self.df.loc[:, "doc_total_per_ask_long_range_mean"] = doc_total_per_ask_long_range_mean
         self.df.loc[:, "doc_total_per_ask_mean"] = doc_total_per_ask_mean
 
+        # All energy-related costs per ASK (energy + carbon tax - subsidy + energy tax)
```
antSGS, line 2273: Could you use a name close to the convention with DOC please? and maybe use 'net' instead of 'all' to make explicit the inclusion of taxes/subsidies ? 

like :  doc_net_energy_per_ask_mean

aeromaps/models/impacts/costs/airlines/direct_operating_costs.py
```diff
@@ -2264,6 +2285,22 @@ def compute(
...
+            - doc_energy_subsidy_per_ask_mean
+            + doc_energy_tax_per_ask_mean
+        )
+
+        # Convert to per-RPK using ASK-weighted costs over total RPK
+        total_ask = ask_long_range + ask_medium_range + ask_short_range
+        total_rpk = rpk_long_range + rpk_medium_range + rpk_short_range
+        doc_all_energy_costs_per_rpk = (
+            doc_all_energy_costs_per_ask_mean * total_ask / total_rpk
+        ).where(total_rpk > 0, 0)
```
antSGS: Wouldn't it work with load factor ?

aeromaps/models/impacts/costs/airlines/direct_operating_costs.py
```diff
@@ -2264,6 +2285,22 @@ def compute(
...
+            doc_energy_per_ask_mean
+            + doc_energy_carbon_tax_per_ask_mean
+            - doc_energy_subsidy_per_ask_mean
+            + doc_energy_tax_per_ask_mean
+        )
+
+        # Convert to per-RPK using ASK-weighted costs over total RPK
+        total_ask = ask_long_range + ask_medium_range + ask_short_range
+        total_rpk = rpk_long_range + rpk_medium_range + rpk_short_range
+        doc_all_energy_costs_per_rpk = (
```
antSGS: Same as doc_all_energy_costs_per_ask_mean, i think something like 

doc_net_energy_per_rpk_mean would be better alligned with the naming here

antSGS:
Maybe you can revert notebooks and files modified in tutorials as these are just outputs things?

aeromaps/plots/multi_scenario/__init__.py
```diff
@@ -83,4 +84,5 @@
     "doc_comparison": DOCComparisonPlot,
     "doc_energy_comparison": DOCEnergyComparisonPlot,
     "airfare_comparison": AirfareComparisonPlot,
+    "all_energy_costs_per_rpk_comparison": AllEnergyCostsPerRPKComparisonPlot,
```
antSGS: Same as in the model 'doc_net' instead of all energy maybe?

aeromaps/plots/single_scenario/__init__.py
```diff
@@ -126,11 +128,13 @@
     "doc_fleet_breakdown": DOCEvolutionBreakdown,
     "doc_fleet_category": DOCEvolutionCategory,
     "airfare_breakdown": AirfareEvolutionBreakdown,
+    "all_energy_costs_per_rpk_breakdown": AllEnergyCostsPerRPKBreakdown,
```
antSGS: Same, 'doc_net_per_rpk_breakdown' ? 


Also, could you clarify the role of this plot compared to doc_fleet_breakdown ?

aeromaps/plots/single_scenario/costs.py
```diff
@@ -532,6 +532,299 @@ def _update_plot_elements(self):
...
+        so that the stacked areas are consistent with ``doc_all_energy_costs_per_rpk``
+        (which is derived from per-ASK costs × total_ASK / total_RPK and therefore
+        excludes freight energy).
+        """
+        y = self.prospective_years
+
+        # Pre-compute passenger-to-total energy ratios per aircraft type so
+        # that each pathway's energy is scaled to passenger-only.
+        pax_ratios = {}
+        if self.pathways_manager is not None:
```
antSGS: If it's using pathways manager, maybe it should go in costs_generic

aeromaps/plots/single_scenario/costs.py
```diff
@@ -532,6 +532,299 @@ def _update_plot_elements(self):
...
+                "carbon_tax": np.where(total_rpk > 0, energy
+                * self.df.loc[y, f"{p.name}_mean_unit_carbon_tax"].fillna(0).values
+                / total_rpk, 0),
+                "tax": np.where(total_rpk > 0, energy
+                * self.df.loc[y, f"{p.name}_mean_unit_tax"].fillna(0).values
+                / total_rpk, 0),
+                "subsidy": np.where(total_rpk > 0, energy
+                * self.df.loc[y, f"{p.name}_mean_unit_subsidy"].fillna(0).values
+                / total_rpk, 0),
+            }
```
antSGS: Not sure i have the logic right here. If an aircraft uses hydrogen, 'ratio' won't be null with dropin fuel in the loop?

aeromaps/plots/single_scenario/energy_mix.py
```diff
@@ -157,16 +157,86 @@ def _update_plot_elements(self):
 ]
 
 
+# ---------------------------------------------------------------------------
+# Blending mandates — relative share of drop-in fuel by energy origin (%)
```
antSGS, line 161: fuel_shares plot already show fuel shares by pathways. What does this one adds?

antSGS (changes requested):
Thanks for the PR. Few points i would like clarifications. Also could you highlight what modifications you've made to plots, that were rather in the perimeter of the previous PR. And the way couplings are initialized looks interesting but do we need to shift to that solution for previous applications as well?

aeromaps/models/air_transport/air_traffic/rpk_constant_elasticity.py
thomasplanes: +1. An alternative is to generate a unique RPK model, but to specify an option regarding the chosen one (e.g. exogenous, Logistic GDP...). I have a similar issue with another model related to climate impacts (effects of fuel on contrails)

aeromaps/core/process.py
thomasplanes: Good question. I think that it has no impact for the user, I tried with a tutorial notebook. But it is weird that no mistake appears due to the fact that this vector is not in the default parameters.json

thomasplanes (commented):
Thank you for the PR. In addition to my direct comments, here are a few remarks:
- Some tests failed, probably due to some changes in the core
- For the notebook "model_comparison", you should take 2 clearer names, such as e.g. Logistic model - With/Without price elasticity, it is not clear currently for the user. In addition, a comparison with the non-logistic functions for RPK could be of interest. Finally, a comparison with Antoine's models could be relevant, for comparing the elastricity values (airfare vs energy consumption price)
- For the notebook "scenario_comparison", I think you should add the reference RPK from SSP/RCP on the fist plot. For the other conventional plot of CO2 emissions, the baseline 2019 technology emissions (in red) is not relevant here. You could add an option to the plot for removing this line if required
- You have made some changes to the tutorials, why?
- You created some tests only for WCTR, is it relevant?

aeromaps/core/gemseo.py
iancostalves: check the usage in the new files added to models.air_transport
models just override the _initialize_df() method by applying the regular method from AeroMAPSProcess, and then defining the coupling defaults, if coupling defaults are assigned, they are used instead of what is in the parameters
personally, this makes more sense to me instead of assigning values directly in parameters (which in the end won't be the real values used in the simulation once the MDA is converged), but I can revert it back

aeromaps/core/process.py
iancostalves: No, if you check the logic of the for loop (left as it was before) you'll see that it loops over parameters (may not contain population_init, gdp_init) and only verifies if it found any item with the names in list_init
if it was looping over list_init and then fetching values from parameters, then they would be mandatory.
I simply kept your own logic and added the new entries for the names of values that can allow vector inputs (which are not mandatory)

aeromaps/core/process.py
iancostalves: This was a functionality I have just added so that plots can also accept figure and ax objects instead of making a new figure every time. This is very practical for making graphs with a zoom-in (check the RPK graphs in the new notebooks).

aeromaps/plots/single_scenario/energy_mix.py
iancostalves: It adds a stackplot instead of lines, which IMO makes it easier to understand the complementarity of shares.

aeromaps/models/air_transport/air_traffic/rpk_constant_elasticity.py
iancostalves: I have renamed model classes and files into:
- price_elasticity.py (as before), containing RPKPriceElasticity (RPKWithElasticity before)
- price_and_income_elasticity.py containing RPKPriceIncomeElasticity
- price_elasticity_logistic_income.py containing RPKLogisticIncomePriceElasticity
does this work for you?

The issue with specifying a unique model is that each model uses different inputs (couplings) from other disciplines (no prices for the exogenous growth, airfare for the simple price elasticity, doc_energy for these new 2 models) meaning the assembled model would have to declare all these as dependencies even if they are not used in the calculation (making extra couplings which may not be needed). By the way, this is also why I added tests for these 2 new configurations to ensure they all work with the rest of AeroMAPS.

aeromaps/models/impacts/costs/airlines/direct_operating_costs.py
iancostalves: Good catch, better than to import ask and rpk from all segments, I have switched to using only load factor. But we might have to review this later if they are defined per market instead of globally...

aeromaps/plots/single_scenario/__init__.py
iancostalves: Renamed for consistency. As the name suggests, the plot is similar to doc_fleet_breakdown, but it keeps only the energy expenses (the non_energy part is excluded), this is mainly because the non_energy part is not included in the elasticity since my calibrations only have energy prices, and I wanted to compare only this component that was driving the elasticity.

aeromaps/plots/single_scenario/costs.py
iancostalves: I don't understood the issue you're highligting here. If there are more aircraft types, then pax_ratios is a dict[aircraft_type: float between 0-1], raw_energy is still the total production of the pathway consumed by aircraft_type (including freight), and energy is just the scaling from total to passenger-only.

iancostalves:
I have modified files according to feedbacks. Here are some extra precisions regarding the changes:
- The plots modification was added so that more options can be passed as input to current plots (fig, ax, legend), BUT doing this revealed some issues that were uncatched in the last PR (some plots did not inherit the base class, adding new plots had to be updated with adding new tests), these are fixed now
- The tutorials changed are the only ones that have notebook tests (verifying that reference outputs are the same as obtained outputs), the issue is that whenever a new model output is added these reference files have to be updated. In my opinion, these notebook tests should be removed or replaced by other unit tests. I could make these new tests myself, but for that I'd need to know what was the goal of adding these tests in the first place?
- The model_comparison notebook was modified to have more explicit names and to compare with Antoine's models
- The scenario_comparison notebook was also modified, BUT not to show the RPK reference for the SSP/RCP. The scenarios I'm using (original SSPs) did not model aviation apart from other transportation so there is no reference values to compare to. Therefore, I prefered to just make a final plot in the end where the RPK values I obtain (my model with AR6 input) are compared with RPK values from other more recent scenarios (AR6 output). There are still issues with this because some models don't have the same aviation scope (international, commercial, ...).
- The models I have added are directly in core.models so that they can be used by anyone using AeroMAPS (not just for this publication). Ideally, all model combinations should be tested to ensure they are consistent with eachother. Currently there are too many models and this would be prohibitive, but I think the best practice would be for new models to systematically add tests that verify compatibility with the rest

iancostalves:
The scenario comparison with only line plots was difficult to compare many scenarios, so I also added options to make "envelopes" according to how scenarios are grouped. These can also be used for visualizing uncertainty @felixpollet , in case you want to use it I recommend taking a look at the tutorial on comparing scenarios (updated basic usage), or the wctr notebook (more advanced usage, with plots with zoom-in).

These plot changes are in the scope of the previous PR on scenario comparison, but some bugs and extra functionalities were only detected when applying features to this notebook.

aeromaps/core/gemseo.py
ScottDelbecq: Interesting but `which in the end won't be the real values used in the simulation once the MDA is converged` is also the case for the values you provide in `_coupling_defaults`, no?

aeromaps/tests/plots/test_single_scenario_plots.py
```diff
@@ -76,328 +88,23 @@ def process_simple():
...
-
-def test_energy_consumption_plot(process):
-    """Test that energy_consumption plot can be created."""
-    plot = process.plot("energy_consumption", save=False)
-    assert plot is not None
 
+@pytest.mark.parametrize("plot_name", _STANDARD_PLOTS)
+def test_standard_plot(process, plot_name):
+    """Every plot reachable through the full-config process must build."""
+    assert process.plot(plot_name, save=False) is not None
```
ScottDelbecq, line 100: 😍

ScottDelbecq (approved):

aeromaps/core/gemseo.py
antSGS: Fine for me. @ScottDelbecq i think the idea is that it's not important to have an overview on these. I'm jsut wondering if there are any situations in which the coupling default value could have an influence on the convergence? In that case it isn't easy to modify the value without opening the hood of aeromaps ?

aeromaps/core/process.py
antSGS: cool

aeromaps/core/process.py
antSGS: You're right, got it!

aeromaps/core/process.py
```diff
@@ -374,9 +374,14 @@ def setup_mda(self):
         # Initialize disciplines
         self._initialize_disciplines()
 
+        # Tolerance must be tight enough to resolve the price-elastic demand loop
```
antSGS, line 399: I've opened a todo in the other branch there to get these params as kwargs of the setup mda to be able to set them without opening the core. 
We just have to think of keeping the todo once merging.

aeromaps/models/air_transport/air_traffic/price_and_income_elasticity.py
antSGS: cool, got it now.

aeromaps/models/impacts/costs/airlines/direct_operating_costs.py
antSGS: Thanks

aeromaps/models/air_transport/air_traffic/rpk_constant_elasticity.py
antSGS: Good for me

aeromaps/plots/single_scenario/__init__.py
antSGS: ok fine

aeromaps/plots/multi_scenario/__init__.py
antSGS: still the "all_energy" naming but that's no big deal!

aeromaps/plots/single_scenario/costs.py
antSGS: Thanks :)

aeromaps/plots/single_scenario/costs.py
antSGS: I can't understand what i meant anymore... let's call it a day haha

aeromaps/plots/single_scenario/energy_mix.py
antSGS: OK fair enough. Should we use another name? Blending mandate refers rather to a policy than to an actual state i think. "stacked_dropin_fuel_shares" or something like that?

antSGS:
All good for me execpt minor answers you'll find in my previous comments

antSGS (approved):
My status was still on request changes but its now fine. 

Everything went well with the merge? Not too much conflicts ?

iancostalves:
@copilot+claude-sonnet-4.6 resolve the merge conflicts in this pull request, keeping the json files for the joas notebook from this branch instead of the main

PR #144, opened 2026-06-08
New notebook folder for predefined scenarios

### LINKED ISSUE #117: ATAG Waypoint 2050

iancostalves:
This PR adds a new folder for dedicated notebooks on predefined scenarios.

New scenarios added:
- ATAG Waypoint 2nd edition (S1, S2, S3, and variants low/central/high traffic and full lifecycle / CORSIA emissions scopes) and tech-only scenarios (T1-T5 full lifecycle / CORSIA emissions scopes)
- ATAG Waypoint 3rd edition (S1, S2) and tech-only scenarios (T0-T4 full lifecycle / CORSIA emissions scopes)
- Coupled ATAG: similar to ATAG 3rd ed. S1, but with coupled demand-price using WCTR models

Some extra modifications on the code:
- I added a third type of offset (with prescribed yearly volumes) instead of the already present level (only active to match emissions up to) and residual (with a part of current emissions to be offset) ones
- The WCTR demand models were slightly changed to replace estimated traffic between 2020-2024 (covid period), this would need to change if start_year is flexible and is left as a TODO

I marked this PR as draft because there are still some open questions:
- Should I move ICAO LTAG scenarios into this folder? Or is it better not to have multiple copies of config and input files?
- Should I implement the baseline scenarios S0 for 2nd and 3rd editions? Or can the multi-region variant serves as the S0?
- Both the WCTR and the Coupled ATAG scenarios have a custom model for Socioeconomic drivers (GDP, Population), should these be moved into the standard models folder, or be left as custom models only when they are used?


iancostalves:
Updates:
- Moved all ATAG scenario variants into a single folder
- Added 2 new aggregation functions in aeromaps/utils:
    - energy_aggregation: sums a bunch of energy pathways into a single "generic" pathway with the sum of volumes and the volume-weighted mean emission factor
    - single_region: takes in a multi-regional process and returns an equivalent single-region process
- Added a new "light" variant for 3rd edition's S0, S1, S2
    - S0: based on the single-region process equivalent of _a_r_m multi-regional stated policies, but the offsets are simplified
    - S1, S2: based on the full 3rd edition S1, S2 but aggregates biofuel pathways into a single generic biofuel

antSGS:
Hi Ian, 

Thanks for this PR. 

Before starting, we have a pre-commit hook configured to remove the ouputs of .ipynbs in commits ast to focus on actual differences rather than output noise. 

Could you do `pre-commit install` ?

aeromaps/models/impacts/emissions/carbon_offset.py
```diff
@@ -188,6 +188,58 @@ def compute(
         return (residual_carbon_offset_share, residual_carbon_offset)
 
 
+class PrescribedCarbonOffset(AeroMAPSModel):
```
antSGS: What about "ManualCarbonOffset". Sounds more antural to me but its no big deal, your call.

aeromaps/models/impacts/others/others.py
```diff
@@ -97,6 +97,15 @@ class DropinFuelConsumptionLiterPerPax100km(AeroMAPSModel):
     def __init__(self, name="dropin_fuel_consumption_liter_per_pax_100km", *args, **kwargs):
         super().__init__(name=name, *args, **kwargs)
 
+    def _initialize_df(self):
```
antSGS, line 100: Why is that a coupling default now? That is an output of energy_carriers_means.py  f"{aircraft_type}_mean_lhv" which is upstream calculated i think. 
What is your opinion?

antSGS:
Just two minor questions regarding

`energy_aggregation.py`: How does it work, does it writes a new yaml file with the aggregated fuel and runs teh scenario again? 

`single_region.py`: that is a nice feature, just to be sure to keep a clear enough workflow: what is the added value over aggregating teh variables of interest of a multi regional process? Is it to make faster run//to make all plots working? Its a bit vague now in my ahead (so much PR/Features at the same time before the break...) but i thnik i made plots availbale at the global level with a check on the existence of the variables as you had implemented. 

Besides, i reviewed these files asking claude to check them and he found minor bugs i commented below directly on the files

aeromaps/plots/multi_scenario/fuel_supply.py
```diff
@@ -0,0 +1,599 @@
+"""Multi-scenario comparison plots for fuel supply and production."""
+
+from aeromaps.plots.multi_scenario_plot import MultiScenarioPlot
+
+
+class FuelSupplyBreakdownPlot(MultiScenarioPlot):
```
antSGS: Claude catched a potential problem with a merge of main back into your branch with commit bf7062ee on main renamed fuel_supply.py → drop_in_supply.py

Thus some functions are apparently duplicated in both files.

antSGS: Not imported in plots __init__.py either

antSGS: Calude tells me the version in drop_in_supply.py is more advanced than this one. It let you explore this issue i do not want to delete the wrong one...

aeromaps/models/impacts/emissions/carbon_offset.py
antSGS: Should we also bound the offset to zero emissions ? or we allow negative emisisons that way?

aeromaps/utils/energy_aggregation.py
```diff
@@ -0,0 +1,234 @@
...
+                ),
+                "emission_index": emission_index,
+            },
+            "economics": {},
+        },
+        "outputs": None,
+    }
+
+    if mfsp_per_carrier:
+        generic_mfsp = _weighted_mean_series(mfsp_per_carrier, quantity_per_carrier)
```
antSGS: it could fail if a cariiers has no MFSP imo and divide lets say 3 cariiers mfsp by 5 cariiers quantities?

aeromaps/utils/energy_aggregation.py
```diff
@@ -0,0 +1,234 @@
...
+            _series_to_yearly(environmental["co2_emission_factor_without_resource"], years)
+        )
+        if emission_index is None:
+            emission_index = environmental.get("emission_index")
+
+        economics = inputs.get("economics") or {}
+        if "mean_mfsp_without_resource" in economics:
+            mfsp_per_carrier.append(
+                _series_to_yearly(economics["mean_mfsp_without_resource"], years)
+            )
```
antSGS: I think we are not covering the whole mfsp and ef here. without ressource is only a part of the mfsp normally, summed to that linked to processes and ressources.

aeromaps/models/impacts/others/others.py
iancostalves: ATAG is made based on quantity mandates, not shares. Since RPK varies in the demand-price coupling, the shares of each dropin fuel are variable so the drop-in fuel LHV is identified as gemseo as a strong coupling here.
Because the values configured in AeroMAPS for the LHV of drop-in is all the same, the coupling is active but this has no practical effect since its values are never changed.

aeromaps/utils/energy_aggregation.py
```diff
@@ -0,0 +1,234 @@
...
+
+        technical = inputs["technical"]
+        rsc_map = technical.get("resource_specific_consumption", {})
+        # A biomass pathway consumes a single feedstock; take its scalar value.
+        rsc_per_carrier.append(float(next(iter(rsc_map.values()), 0.0)))
```
antSGS: we could have several ressources per pathways i think (e.g. electrofule with h2 and co2)

aeromaps/utils/energy_aggregation.py
```diff
@@ -0,0 +1,234 @@
...
+        quantity = _series_to_yearly(mandate["mandate_quantity"], years)
+        quantity_per_carrier.append(quantity)
+        total_quantity_weight.append(float(np.sum(quantity)))
+
+        environmental = inputs["environmental"]
+        ef_per_carrier.append(
+            _series_to_yearly(environmental["co2_emission_factor_without_resource"], years)
+        )
+        if emission_index is None:
+            emission_index = environmental.get("emission_index")
```
antSGS: it's only th eindex of the first/last pathways merge here? Why do we not use the weighted average?

aeromaps/utils/single_region.py
```diff
@@ -0,0 +1,287 @@
...
+
+    for region_id, region_config_path in region_configs.items():
+        process = create_process(configuration_file=region_config_path)
+        process.compute()
+        df = process.data["vector_outputs"]
+        idx = df.index
+
+        def at_years(series):
+            # Restrict/reindex the model series onto the year grid.
+            return np.array([float(series.loc[y]) if y in idx else 0.0 for y in years])
```
antSGS, line 231: is that okay? I mean if a region goes upt to a given year then fuel mandates go to 0 for instance? Shouldn't we use a warning therE?

antSGS: Claude: **1. Zero-fill outside the model index** (L112). `at_years` returns `0.0` for any year not in `df.index` instead of failing. If `year_range` reaches past a region's `end_year`, those years get zero SAF *and* zero drop-in, and the `where=total_dropin > 0` guard (L130) turns that into a silent **0% mandate tail** — indistinguishable from a deliberate policy. `year_range` defaults to `(2020, 2050)` regardless of what the regions actually cover, so this is easy to hit by accident. Suggest raising when `years` isn't contained in `idx`.

aeromaps/utils/single_region.py
```diff
@@ -0,0 +1,287 @@
...
+
+        total_dropin += at_years(df["energy_consumption_dropin_fuel"])
+        if saf_var in df.columns:
+            saf = at_years(df[saf_var])
+            ef = at_years(df[ef_var]) if ef_var in df.columns else np.zeros(len(years))
+            total_saf += saf
+            total_saf_ef += saf * ef
+            if reference_region is None:
```
antSGS: Claude: 
**2. All-NaN emission factor reaching the YAML** (L115–121 vs L132–135). `reference_region` is set as soon as the SAF *column exists*, but `global_ef` is NaN wherever `total_saf == 0`. A region declaring the carrier with an all-zero trajectory passes the L123 check while leaving `global_ef` entirely NaN — `ffill().bfill()` has nothing to propagate from — and `_custom_data_type` writes `float('nan')` straight into the generated energy file. Gating on `saf.sum() > 0` rather than column presence catches it at the source.

aeromaps/utils/single_region.py
```diff
@@ -0,0 +1,287 @@
...
+        idx = df.index
+
+        def at_years(series):
+            # Restrict/reindex the model series onto the year grid.
+            return np.array([float(series.loc[y]) if y in idx else 0.0 for y in years])
+
+        total_dropin += at_years(df["energy_consumption_dropin_fuel"])
+        if saf_var in df.columns:
+            saf = at_years(df[saf_var])
+            ef = at_years(df[ef_var]) if ef_var in df.columns else np.zeros(len(years))
```
antSGS: Claude: substitutes `np.zeros` when `ef_var` is missing, which drags the weighted mean toward zero instead of flagging the gap.

aeromaps/utils/single_region.py
```diff
@@ -0,0 +1,287 @@
...
+def aggregate_regions_to_single_process(
+    configuration_file,
+    output_config,
+    output_energy_file,
+    demand_override=None,
+    standards_override=None,
+    fuel_carrier="generic_saf",
+    reference_region=None,
+    output_json=None,
+    year_range=(2020, 2050),
```
antSGS: hardcoded instead of attributes ? Possible to use simulation_start_year and end_year?

aeromaps/models/impacts/others/others.py
```diff
@@ -97,6 +97,15 @@ class DropinFuelConsumptionLiterPerPax100km(AeroMAPSModel):
...
+        super()._initialize_df()
+        self._coupling_defaults = {
+            "dropin_fuel_mean_lhv": pd.Series(
+                44.0,  # MJ/liter
```
antSGS: i guess this value is MJ/kg right?

antSGS: SAme below in the unit

antSGS (changes requested):
Hi Ian, thanks for the PR, few changes are required befor merging. 
Good job!

iancostalves:
Hey, if you check the ATAG scenario folders you'll se that there is a "light" version of ATAG scenarios made to simplify their incorporation into the web-app.
- Energy aggregation just creates the yaml file somewhere, the configuration still has to be imported in a process and then run. The light S1 and S2 scenarios are based on the "full" scenarios, but they aggregate all the biofuels into a single generic biofuel with mean emission factor (easier to port to not handle too many biofuels).
- Single region: creates the process directly from the aggregation of a multi-region scenario. Since the ATAG S0 is based on this multi-region to compute the offsets and SAF, I simply make an equivalent single-region process that can be fed directly into the web-app.

aeromaps/utils/energy_aggregation.py
iancostalves: In this case (2/5 without MFSP) would it be a problem if they are considered zero ?

aeromaps/utils/single_region.py
iancostalves: I'd leave it as is, the most common case where this is turned on is the case where mandates start later than the prospection year so it fills with zero before any ramp-up (almost all scenarios fill mandates/quantities until the last year). I could raise a warning if you want, but IMO there are already too many of them everywhere.

aeromaps/models/impacts/generic_energy_model/common/energy_carriers_manager.py
```diff
@@ -131,3 +131,57 @@ def get_all_types(self, parameter: str) -> List:
                 if getattr(carrier, parameter, None) is not None
             }
         )
+
+
+def build_pathways_manager(energy_carriers_data, energy_processes_data=None):
```
ScottDelbecq, line 136: Good initiative! This is something that should become more general so I opened an issue #160.

ScottDelbecq (changes requested):
Thanks @iancostalves for this significant work!
I would just like that data files related to reference scenarios live in a folder `scenarios` in the `resources` folder. 
For the notebooks using them we can execute in the first cells to copy them to a `workdir` so the user has a real sandbox.

ScottDelbecq (approved):

antSGS (commented):
> `build_pathways_manager` looks up each process's consumed resources using `f"{name}_resource_names"`, but the processes YAML (e.g. `aeromaps/resources/data/default_energy_carriers/processes_data.yaml`) declares the key as `resource_names`. As written, this will almost always fall back to `[None]`, so plots/metadata that rely on `resources_used_processes` lose the actual resource mapping. 

--> wrong IMO, copilot does not know that the string is stripped out before being read in the discipline
--> Verified no issue , copilot missed the way self.energy_processes_data was built

antSGS (approved):
All good for me now

Issue #142, opened 2026-05-26
Freight bottom-up fleet support


antSGS:
## Context

The user-defined markets refactor
([REFACTOR_MARKETS_PLAN.md](REFACTOR_MARKETS_PLAN.md)) shipped two freight
efficiency disciplines, both opt-in via the user's model dict in
[core/models.py](aeromaps/core/models.py) (same convention as the passenger
``simple_shares`` / ``complex`` split):

- [`FreightAircraftEfficiency`](aeromaps/models/air_transport/aircraft_fleet_and_operations/fleet/aircraft_efficiency.py#L560)
  — drop-in efficiency and propulsion mix derived from the passenger fleet.
  Included by default in `models_efficiency_top_down`,
  `models_efficiency_top_down_interp`, and `models_efficiency_bottom_up`.
- [`FreightAircraftEfficiencySimple`](aeromaps/models/air_transport/aircraft_fleet_and_operations/fleet/aircraft_efficiency.py)
  — per-freight-market drop-in gain curve, no alternative propulsion. Swap it
  into a model dict in place of `FreightAircraftEfficiency` when desired;
  requires each freight market to carry an `inputs.efficiency_simple` block
  (defaults available under `defaults.freight` in `markets.yaml`).

What is **not** supported yet: a freight market driven by its own fleet of
freight aircraft (B777F-class) with S-curve assignment — the bottom-up
equivalent of the passenger
[`FleetModel`](aeromaps/models/air_transport/aircraft_fleet_and_operations/fleet/fleet_model.py).
This issue tracks that work.

## Why it's a deep refactor

The fleet layer was explicitly built passenger-only in Phase 3:

- `traffic_type="passenger"` filters at
  [fleet_model.py:791](aeromaps/models/air_transport/aircraft_fleet_and_operations/fleet/fleet_model.py#L791),
  [:959](aeromaps/models/air_transport/aircraft_fleet_and_operations/fleet/fleet_model.py#L959),
  [fleet_numeric.py:68](aeromaps/models/air_transport/aircraft_fleet_and_operations/fleet/fleet_numeric.py#L68).
- DataFrame columns and outputs use `ask_<market>` / `energy_per_ask_<market>_<carrier>`
  templates; no RTK path exists through `FleetModel` /
  `FleetEvolution` / `FleetPerformanceMixin`.
- `aircraft.energy_per_ask` is the per-aircraft attribute; there is no
  `energy_per_rtk` analogue.
[...]

PR #145, opened 2026-06-08
Refactoring of markets in aeromaps for a fully generic yaml description + interactions with fleet

antSGS:
Very large PR with a lot of modifications. 

Everything is synthetized in the title, here's a few more details of waht's included. 
I've let my master working plan in the files #REFACTOR_MARKET_PLAN.MD which can be read for a deeper understanting of the process. 

# User-defined markets refactor + demand-model integration

## Summary

Replaces the hard-coded 4-market fleet structure (`short_range` / `medium_range` / `long_range` / `freight`) with **user-defined markets declared in YAML**, following the energy-carriers pattern. The *market* becomes the top-level grouping concept; subcategories and aircraft stay fleet-internal and are never propagated to downstream models. Downstream disciplines are wired dynamically through `AeroMAPSCustomModelWrapper`, with one GEMSEO discipline instance per market.

This PR also integrates the demand models brought in from the `wctr2026` merge (Ian's price/income and logistic-income elasticity models), migrating them onto the generic-market definition.

## What's included

**Core data model & registry**
- New `Market` dataclass + `MarketManager` registry with `from_yaml` loader (`aeromaps/models/air_transport/markets/`).
- Default `markets.yaml` ships 4 markets reproducing legacy numerics exactly; `models.markets` is opt-in in `config.yaml`, so legacy configs stay functional.
- `_initialize_markets()` in `process.py` flattens each market's `inputs` into `self.parameters` under market-as-prefix names (`<market_id>_<leaf_key>`).

**Air traffic, fleet & downstream models migrated to per-market generic I/O**
- RPK / RTK / ASK / load-factor models now iterate over the registry (one discipline per market) via a new `markets_factory.py`.
- Fleet model: generic calibration loop, data-driven energy-type handling, vectorized performance methods; `fleet.yaml` switched to `markets:` / `market_served:`.
- Downstream chains (energy consumption, DOC, CO₂ / non-CO₂ emissions, abatement cost, total airline cost) migrated to the custom-wrapper pattern with dynamic market-templated names.
- New simplified freight efficiency model.

**Demand-model integration (recent merge edits) --> @iancostalves I need a very cautious review//edits from yourself on that as it's your models i've never worked with.**
- `RPKPriceIncomeElasticity` and `RPKLogisticIncomePriceElasticity` migrated to the generic-market definition; selected via config, with `RPKElasticity` owning the unsuffixed `rpk_<market>` name (base RPK model emits `_no_elasticity`).
- New `markets_factory.py` helpers build per-market RPK/RTK/ASK/load-factor disciplines and the matching aggregators from the registry.

## Variable naming

New market-as-prefix convention (`<market>_<leaf>`, e.g. `short_range_rpk_share_2019`, `medium_range_ask`). The legacy three-convention naming is replaced. Full old→new mapping (renames, `rpk_`/`passenger_` drops, freight-prefix, 1→3 per-market splits) is documented in `REFACTOR_MARKETS_PLAN.md §3.5.1`.

## Behavioral notes for reviewers

- **Cleanups, not regressions:** six always-zero `*_hybrid_electric` placeholder outputs removed (hybrid energy was already routed through the drop-in + electric buckets via `hybridization_factor`).
- **Two intended numerical drifts:** a carbon-tax-lowering-offset bug fix (`a3ed83b3`) and a more-correct COVID-window growth-rate computation (`3950aec3`). Both flagged in `§3.5.1 G`.
- All four reference tutorials (`01`, `02`, `03`, `05`) plus `06` climate were re-run; reference `outputs.json` updated accordingly.

## Out of scope / follow-ups

- Plots & GUI rework beyond the renames already applied ? 
- A true `*_hybrid_electric` output series (genuine feature gap, see `§3.5.1 F` — separate issue).
- migration of WCTR notebooks to work with the generic model (simple theoretically, just to decalere a maket_file.yaml). We can do that together @iancostalves .


antSGS:
Please note that this is a DRAFT PR, heavy changes following the merge with mainrequire following investigations and work (e.g. WCTR notebooks compatibility).

aeromaps/models/air_transport/air_traffic/price_elasticity_logistic_income.py
```diff
@@ -232,137 +189,77 @@ def compute(
...
-        rpk_no_price_total = population * rpk_per_capita_trend
-
-        rpk_no_elast_short = rpk_no_price_total * short_range_rpk_share_2019 / 100
-        rpk_no_elast_medium = rpk_no_price_total * medium_range_rpk_share_2019 / 100
-        rpk_no_elast_long = rpk_no_price_total * long_range_rpk_share_2019 / 100
+        # --- Per-market split (historic uses rpk_init * share), measures, and totals ---
+        n = self.end_year - self.prospection_start_year
+        base_year = self.prospection_start_year - 1
+        output_data = {}
+        rpk = rpk_no_elasticity = rpk_model_without_covid = rpk_reference = None
```
iancostalves: why is this done here?

aeromaps/models/air_transport/air_traffic/price_elasticity_logistic_income.py
```diff
@@ -232,137 +189,77 @@ def compute(
...
+            rpk_no_elasticity = (
+                rpk_m_no_elast if rpk_no_elasticity is None else rpk_no_elasticity + rpk_m_no_elast
+            )
+            rpk_model_without_covid = (
+                rpk_m_wc if rpk_model_without_covid is None else rpk_model_without_covid + rpk_m_wc
+            )
+            rpk_reference_m = self._full_series(input_data[f"rpk_reference_{mid}"], 0.0)
+            rpk_reference = (
+                rpk_reference_m if rpk_reference is None else rpk_reference + rpk_reference_m
+            )
```
iancostalves: why recompute aggregated values per market? I get that measures can have a per-market impact and lead to the total rpk to be different than the sum of share_m * rpk_m, but then instead of recomputing it all you can just get the measures impact as sum of share_m * rpk_m * (1 - measures_impact) and compute rpk in the end, no? In the end it should reduce the amount of computations made

aeromaps/models/air_transport/air_traffic/price_elasticity_logistic_income.py
```diff
@@ -68,131 +133,23 @@ def _apply_price_delay(self, price):
...
+
+        The global per-capita income trend is multiplied by the price index; the result is
+        split across passenger markets by their 2019 RPK share, multiplied by each market's
+        measures impact and summed into the total ``rpk``. Historic years are pinned to the
+        exogenous ``rpk_init`` split.
         """
+        rpk_init = input_data["rpk_init"]
+        population = input_data["population"]
+        gdp_per_capita = input_data["gdp_per_capita"]
+        doc_net_energy_per_rpk_mean = input_data["doc_net_energy_per_rpk_mean"]
```
iancostalves, line 155: market-averaged price used for demand feedback (same as the current WCTR version), but since the markets are already being refactored, could we have an optional flag to the market definitions (extra parameter to be put if demand-price coupling is on) that allows to choose wether to use averaged or market-specific price? it's OK if this is out of scope for now

aeromaps/models/air_transport/air_traffic/price_and_income_elasticity.py
```diff
@@ -34,6 +53,39 @@ def __init__(self, name="rpk_constant_elasticity", *args, **kwargs):
...
+            "doc_net_energy_per_rpk_mean": pd.Series([0.0]),
+            "gdp_per_capita_2019": 0.0,
+            "gdp_per_capita_covid_end": 0.0,
+            "gdp_per_capita_init": pd.Series([0.0]),
+            "population_init": pd.Series([0.0]),
+        }
+        for mid in self.passenger_market_ids:
+            self.input_names[f"{mid}_rpk_share_2019"] = 0.0
+            self.input_names[f"rpk_{mid}_measures_impact"] = pd.Series([0.0])
+            self.input_names[f"rpk_reference_{mid}"] = pd.Series([0.0])
```
iancostalves, line 77: So essentially one big discipline for all markets, might me good to document this properly. In the refactor plan it's mentioned that one discipline per market is used, but I guess this is not true everywhere (maybe explicit that its only for the RPK/load factor/ASK).

iancostalves (commented):
Big PR, I liked how well documented REFACTOR_MARKETS_PLAN.md is, and the step by step is really helpfull for understanding the workflow.

As this PR is still a draft, I only did a draft review (read REFACTOR_MARKETS_PLAN.md and the WCTR files that concern me), but I already have a few of questions regarding some design decisions:
- Coupled-demand models: modifications seem OK to me, most comments made are minor performance stuff or general questions on price-demand coupling (do we allow them to be per-market/per-region or just keep it at the global level?)
- Per market vs per field disciplines: precise somewhere in the description of the models if they are computed per-market or across all markets. Especially when multimarket+multiregion this can be cumbersome to handle
- Generic genericity: with energy/regions/markets disaggregation seems to be multiplying and handling the namespaces/separators can get confusing, maybe we could make a base class for this? Ex: Model aggregator, qui definit le separateur ( "_" en ce moment mais possible avec "." ou ":") les entrées comunnes/spécifiques, les sorties agrégés/spécifiques, les sous-classes ont juste à implémenter la fonction base et l'aggregator crée les disciplines selon une strategie qu'on peut choisir (discipline monolithique vs N discs par composant). A discuter...
- Freight: since the markets are already being refactored, why not increase detail in freight models (belly + freighter)? One quick-fix would be to simply allow for more than one freight market, or we could go into more detail expliciting belly-RTK as per-market attribute (alongside RPK and ASK), and a deicated market for freighters with no RPK.

thomasplanes (approved):
Excellent contribution Antoine! No remark on the PR, the data format for the configuration file was already validated

antSGS:
This PR now contains three other chained PR

antSGS:
# Chained PR 3: merge feature_flexible_start_year into fleet_refactoring

Hi everyone, i propose to submit this little PR about flexible start year. Either we can merge into main after the merge of fett-refactoring, or merge it into fleet refactoring.   
Any case I request your review here as both features are quite different.   
I'm a bit busy so i'll kepp it as a draft but testing is not fully complete. If one wants to stress test the branch, feel free to do so while updating the *reference outputs*, but i need to move on right now on a child branch to implement other features for WP1.

Important stuff: 
--> I've uploaded a test notebook in which one can strat with another year and documents how ot do it.  
--> Normally, all notebooks with default years run, but i have not chack one by one, and most importantly ** i have not updated reference outputs yet**. 
--> Claude's summary of the modifications, also documented in Flexible_start_year_plan.md with more details.   


**Source:** `feature_flex_prospection_start_year` → **Target:** `fleet-refactoring`

## Summary

Removes the hard-coded `prospection_start_year = 2020` assumption baked into the
engine through repeated `.loc[2019]` / `.loc[2020]` / `(k - 2019)` constructs and
`*_2019`-named parameters. After this change a user can set
`prospection_start_year` to any year (e.g. 2025) — supplying their own extended
`*_init` vectors and updated reference scalars — and the simulation runs
unchanged. The other bounds (`historic_start_year=2000`,
`climate_historic_start_year=1940`, `end_year=2050`) are untouched.

This is a **code-only** change: defaults stay at 2020 and no new default data is
shipped. For `prospection_start_year == 2020` the math is identical (regression
verified, see below) — the work is parameterization + renaming.

## What changed

### 1. Single source of truth — `last_historical_year`
- `Parameters.last_historical_year` is a **derived `@property`** returning
  `prospection_start_year - 1`. It is computed, never stored, never persisted to
  JSON, so it cannot drift out of sync.
- `AeroMAPSModel` exposes `self.last_historical_year` to every model, so models
  read it instead of the literal `2019`.

### 2. Last-historical-year reads generalized (`.loc[2019]`, `(k - 2019)`)
All "last historic year / calibration anchor" reads now route through
`self.last_historical_year`:
- `load_factor.py`, `aircraft_efficiency.py`, `fleet_model.py`
  (reference-aircraft energy-per-ASK calibration)
- `carbon_offset.py`, `fleet_abatement_cost.py`, `co2_emissions.py`,
  `carbon_budget.py`
- air-traffic models: `rpk_market.py`, `rtk_market.py`,
  `short_range_distribution.py` (the `2019` reference-year anchor; intermediate
  waypoints kept), `price_and_income_elasticity.py`,
  `price_elasticity_logistic_income.py`

### 3. COVID-2020 patches made conditional
The unconditional `self.df.loc[2020, …] = …` COVID overrides in `load_factor.py`
and `aircraft_efficiency.py` are now guarded by `if prospection_start_year <= 2020`,
becoming no-ops when 2020 falls outside the prospective window. User-provided
historic `*_init` data is expected to already reflect the observed 2020 drop.
The `covid_*_2020` parameter **names are kept as-is** — they describe a real
historical event, not a generic pivot — and are simply ignored when the guard
short-circuits.

### 4. `*_2019` parameter rename → `*_last_historical_year`
Parameters whose `_2019` suffix encoded "last historical year" (not a 2019 event)
were renamed, e.g. `gdp_per_capita_2019`, `<mid>_rpk_share_2019`,
`<mid>_energy_share_2019`, `freight_energy_share_2019`,
`carbon_offset_baseline_level_vs_2019*`, `world_co2_emissions_2019`.

**Migration is fail-fast, not silent.** Instead of a backward-compat alias shim,
the loader raises a clear `ValueError` when it sees a stale renamed key
(`RENAMED_INPUTS` exact map + `RENAMED_INPUT_SUFFIXES` for the `_share_2019`
family), telling the user the new name. This keeps configs honest rather than
masking a half-migrated file. All in-repo configs, default `parameters.json` /
`markets.yaml`, and tutorial notebooks were migrated to the new names.

### 5. Input-length validation
`*_init` vector inputs are checked against
`prospection_start_year - historic_start_year` at load time and raise a clear
`ValueError` (offending key, expected vs actual length) instead of an obscure
pandas error.

## New artifacts
- Tutorial: **`change_the_prospection_start_year.ipynb`** with a worked
  `prospection_start_year=2025` config (`inputs_2025.json`, `config.yaml`).
- All notebooks migrated from the `2019` key to `last_historical_year`.

## Notable deviations from the original plan
- **`last_historical_year` is a derived property** rather than an attribute set in
  `_initialize_inputs` — cannot drift, never persisted.
- **No backward-compat alias shim.** The plan proposed silent aliases with a
  deprecation warning; the implementation uses a hard rename guard that errors,
  so stale configs fail loudly with the corrected name.

## Verification
- **Default regression (`01_basic`):** zero diff in `outputs.json`
  (`prospection_start_year == 2020` math unchanged).
- **`pytest aeromaps/`:** model-level regressions covered.
- **Custom 2025 scenario:** loads (Phase E gate passes), `load_factor.loc[2020]`
  comes from the user's historic `rpk_init/ask_init` (not `covid_load_factor_2020`),
  offset baseline reads from `last_historical_year`.

## Out of scope
- Climate historical CSV extension (`temperature_historical_dataset.csv`) — the
  user supplies an extended CSV for `prospection_start_year > 2020` with climate
  models enabled.
- Publication notebooks keep their hard-coded year arrays unless requested.

antSGS:
# Chaine PR 2: merge fleet_paco_models into feature_flet_prospection_start_year

Open PR as draft ==> The way I integrated paco's work is far from definitive/optimal. I went for efficiency first as to land them in AeroMAPS without many modification to its inputs/outputs conventions. In the end, it diverges completely from the AeroMAPS Fleet Model: only market level values are out of Paco's model. Does not work with costs/Nox/... so far. Heavier work to do if we want, with a full integration to conventionnal FleetModel.

The branch is also rebased on private custom_workflow branch to recuperate non private features developped, as detailled below. 


## What this PR does

It adds **Paco's "push" fleet model** as a first-class, optional AeroMAPS model, and bundles
the **shared engine changes** that the `custom_workflow` multi-region branch needs to run. Both
build on the flexible start-year work already in the base branch — nothing in the existing
default scenarios changes; everything here is opt-in and coexists with the current top-down /
bottom-up paths.

Paco's engine works the opposite way to our usual fleet model: instead of choosing market
shares from an S-curve, it starts from **aircraft deliveries** and lets retirement + utilisation
curves decide what fleet survives, then sums the energy bottom-up. AeroMAPS feeds it the traffic
(ASK) and it returns the standard efficiency outputs — so it slots into the rest of the pipeline
transparently. Because Paco keeps editing the engine, we wired its inputs/outputs the AeroMAPS
way rather than wrapping it as a black box.

---

## Part 1 — The push fleet model

### How it's declared / selected

It's a normal selectable efficiency model. In a scenario `config.yaml` you:

1. **Select the engine** in the standards list:
   ```yaml
   models:
     standards:
       - models_traffic
       - models_efficiency_push        # <- the push fleet engine
       - models_energy_without_fuel_effect
       - ...
   ```
2. **Point at the markets file** (Paco's TP/RJ/NB/WB passenger segments + freight):
   ```yaml
     markets:
       markets_data_file: "./markets_push.yaml"
   ```
3. *(Optional)* **Override engine input files** via a `models.fleet_push` block, parallel to
   `models.fleet`. Omit it entirely and it falls back to the shipped defaults; set any key to
   `default` for the shipped resource, or give a path to override just that file:
   ```yaml
     fleet_push:
       classification_data_file: default
       market_param_data_file: default
       in_production_inventory_file: default
       new_aircraft_inventory_file: default
       aircraft_parameters_file: default
       fleet_snapshot_file: default
   ```

After `process.compute()`, the instance is available as a **named handle**:
`process.push_fleet_model` (resolved generically; `None` when the push standard isn't selected).

> **Hard requirement:** the engine is calibrated to the **end-2024 fleet**, so it requires a
> **2024 pivot** (`prospection_start_year = 2025`). `custom_setup` guards this and raises
> otherwise. This is exactly why it depends on the flexible start-year base branch.

### Inputs it works with

- **From AeroMAPS at runtime:** the per-segment traffic `ask_{mid}` (always injected — the
  engine has no internal growth profile) plus the last-historical-year share scalars. The growth
  trajectory itself lives in `markets_push.yaml` (`cagr_reference_periods[_values]` per segment),
  ported from the legacy engine.
- **Static engine data (six files, the `PUSH_FLEET_INPUT_FILE_KEYS`)**, loaded once at setup:
  `classification_data_file`, `market_param_data_file`, `in_production_inventory_file`,
  `new_aircraft_inventory_file`, `aircraft_parameters_file`, `fleet_snapshot_file`. Shipped
  under `resources/data/default_fleet_push/` (+ two calibration workbooks).

### Outputs it emits (the bridge contract)

Two bridges, identical in shape to existing models so downstream disciplines don't notice the
swap:

- **(a) Efficiency bridge** (like `PassengerAircraftEfficiencySimpleShares`): per market `mid`
  and energy type `et` — `energy_per_ask_without_operations_{mid}_{et}` (MJ/ASK) and
  `ask_{mid}_{et}_share`. **Drop-in fuel only** in v1: `dropin_fuel` carries 100 %;
  hydrogen/electric shares are 0 with their energy columns mirroring drop-in so downstream reads
  stay finite.
- **(b) Fleet bridge** (like `SimpleFleetCount`): `"{market}: Aircraft In Fleet"`,
  `{mid}:{type}:aircraft_in_fleet`, and `{mid}:{type}:aircraft_deliveries` — the deliveries
  column being the push model's distinguishing output.

### How to get the plots

The diagnostic charts are a **method on the model**, not registry plots (the retirement /
retirement-age charts need the engine's internal age-resolved arrays that the flat output
columns sum away):

```python
process.compute()
process.push_fleet_model.plot()   # per-segment: fleet, ASK, deliveries, energy,
                                  # retirements, retirement age, energy intensity
```

Call it after `compute()` (it reads the engine arrays cached on the last run; raises otherwise).
Full demo: `notebooks/tutorials/14_push_fleet_model/` (pivot 2025, `end_year` 2070).

---

## Part 2 — Shared changes for the `custom_workflow` branch

These are additive, engine-side changes that live here but exist to support the multi-region
Excel scenario maintained on the **`custom_workflow`** branch (8 regions × 7 markets × 4 aircraft
generations + global freight, run as one `MultiRegionalProcess`). All are gated on new opt-in
fields, so default scenarios are unchanged.

- **Per-year energy improvement** — `continuous_improvement_factor_energy`, an optional per-year
  multiplier on the resolved `energy_per_ask` (in `fleet_model.py` / `fleet_performance.py`).
  Now also on the reference aircraft, so every generation can improve. Absent → no-op.
- **Share-decoupling mode** — a per-year `share` series on a card takes the fleet off the S-curve:
  shares are written directly and base-year reference calibration is skipped
  (`fleet_model.py` / `fleet_assignment.py`). This is what lets the workbook feed all four
  generation shares explicitly.
- **Time-varying fleet productivity** — `ask_year` (in `fleet_numeric.py`) now accepts a per-year
  series, not just a scalar, so aircraft-count productivity can ramp over time.
- **Selectable per-market load factor model** — `create_market_load_factor_models` gains a
  `load_factor_model` switch: `"quadratic"` (default, unchanged) or `"simple_interpolation"`
  (piece-wise linear across workbook waypoints).
- **New standards groups** (`core/models.py`) — `models_efficiency_bottom_up_simple_freight`
  (bottom-up passenger + simple global freight) and `models_fleet_numeric` (standalone fleet
  counts without the flaky cost models).
- **Elasticity start-year fix** (`rpk_market.py`) — the price-elasticity start is now clamped to
  never begin before `prospection_start_year`, which previously crashed for late start years.

---

## Status

- Both demos run end-to-end; core tests stay green (default scenarios untouched).
- Still owed before merge: a full `pytest aeromaps/` re-run to confirm zero regression.

## Where to look

- **Push model:** `FLEET_PUSH_INTEGRATION_PLAN.md` + `notebooks/tutorials/14_push_fleet_model/`
- **Multi-region / custom_workflow:** `PHASE_INTERFACE_SUMMARY.md` +
  `notebooks/custom_workflow/11_custom_markets_multi_regions/`

antSGS:
# Chained PR 3: merge custom_workflow_publipickup into fleet_paco_models

## Summary

Brings the public portion of the private `custom_workflow` branch into `fleet_paco_models`.
Everything under `aeromaps/notebooks/custom_workflow/`, the Docker/CI files for the private
notebook image, and internal planning docs are excluded by construction.

Main features picked up:
- Market shares validation: RPK/RTK/energy shares must sum to 100 % (hard error in `AeroMAPSProcess`)
- Multi-regional aggregation extended to climate and float outputs; vector accessors renamed `get_*_vector_outputs`
- Global-level (aggregation-safe) plotting on `MultiRegionalProcess`, and `write_excel` extended with per-region climate/floats + overall sheets
- Fleet plot rebuilt from `aircraft_share` via stackplot (works in decoupled mode) + performance-contribution plot fix for continuous-improvement factors
- Fixes: null freight share in the push model, elasticity/`_a_r_m` notebook variables
- 27 new tests (aircraft perf-mode validation, multi-regional aggregation, global plotting)

## How it was built

`custom_workflow` forked from an older `fleet_paco_models` and was reconciled later through
internal merges; most of that era's public work had already been cherry-picked into
`fleet_paco_models`. This PR therefore contains:

- **1 squash commit** (`71b8e2b2`) with the net public residual of the pre-reconciliation era
  (original source SHAs listed in the commit message)
- **6 individual cherry-picks** of the post-reconciliation commits, each carrying a
  `(cherry picked from commit …)` provenance line

## Verification

- Public-path tree identity: `git diff custom_workflow <branch>` restricted to public paths is empty
- No commit in this PR touches any private path (audited over the full range history, including the private-folder rename)
- Full test suite green: `pytest tests/` → 100 passed (includes the 27 new tests)

## Merge notes

First step of the integration cascade
`custom_workflow_public_pickup` → `fleet_paco_models` → `feature_flex_prospection_start_year` → `fleet-refactoring` → `main`.
Merge with a **merge commit** (no squash/rebase) to keep the stack fast-forward-clean.

aeromaps/models/air_transport/air_traffic/price_elasticity_logistic_income.py
antSGS: Yes, we can do this, except from the rpk themselves that are really needed per market downstream

aeromaps/models/air_transport/air_traffic/price_elasticity_logistic_income.py
antSGS: Good idea for future developements. However, i need to move on with this PR but we can keep track of this in an enhancement issue?

aeromaps/models/air_transport/air_traffic/price_elasticity_logistic_income.py
antSGS: not very readable you are right and thats a claude way to instanciate the various series. 
Switch to something more readable

aeromaps/models/air_transport/air_traffic/price_and_income_elasticity.py
antSGS: Yes you are right that was the initial plan, but depend on the situation. Here as some operations are made for aggregate values it's maybe a better idea to keep it that way. Otherwise a per discipline + aggregator architecture would be needed and we would create N times the same coupling if i understand it correctly.

antSGS: Added that as a comment

aeromaps/models/air_transport/air_traffic/price_elasticity_logistic_income.py
antSGS: d0d59be1ab0507bbc1ee40dd5a977ac3f2cccf17

aeromaps/models/air_transport/air_traffic/price_elasticity_logistic_income.py
antSGS: d0d59be1ab0507bbc1ee40dd5a977ac3f2cccf17

aeromaps/models/air_transport/air_traffic/price_elasticity_logistic_income.py
antSGS: Done d0d59be1ab0507bbc1ee40dd5a977ac3f2cccf17

aeromaps/models/air_transport/air_traffic/price_and_income_elasticity.py
antSGS: d0d59be1ab0507bbc1ee40dd5a977ac3f2cccf17

antSGS (commented):
> Big PR, I liked how well documented REFACTOR_MARKETS_PLAN.md is, and the step by step is really helpfull for understanding the workflow.
> 
> As this PR is still a draft, I only did a draft review (read REFACTOR_MARKETS_PLAN.md and the WCTR files that concern me), but I already have a few of questions regarding some design decisions:
> - Coupled-demand models: modifications seem OK to me, most comments made are minor performance stuff or general questions on price-demand coupling (do we allow them to be per-market/per-region or just keep it at the global level?)
> - Per market vs per field disciplines: precise somewhere in the description of the models if they are computed per-market or across all markets. Especially when multimarket+multiregion this can be cumbersome to handle
> - Generic genericity: with energy/regions/markets disaggregation seems to be multiplying and handling the namespaces/separators can get confusing, maybe we could make a base class for this? Ex: Model aggregator, qui definit le separateur ( "_" en ce moment mais possible avec "." ou ":") les entrées comunnes/spécifiques, les sorties agrégés/spécifiques, les sous-classes ont juste à implémenter la fonction base et l'aggregator crée les disciplines selon une strategie qu'on peut choisir (discipline monolithique vs N discs par composant). A discuter...
> - Freight: since the markets are already being refactored, why not increase detail in freight models (belly + freighter)? One quick-fix would be to simply allow for more than one freight market, or we could go into more detail expliciting belly-RTK as per-market attribute (alongside RPK and ASK), and a deicated market for freighters with no RPK. 

Issues created for keeping track of your concerns: 

- #151 : generic-genericity
- comment to already opened #142 : freight improvement
- #152 : per amrket elasticities

aeromaps/core/multi_regional_process.py
```diff
@@ -919,44 +1043,163 @@ def plot(
...
+    def _plot_global(
+        self,
+        name: str,
+        save: bool = False,
+        size_inches: Optional[tuple] = None,
+        remove_title: bool = False,
+        fig=None,
+        ax=None,
+        legend=True,
+    ):
```
ScottDelbecq, line 1115:
```suggestion
    # TODO: investigate if plots should be in dedicated class with process (or data) as arg/attribute
    def _plot_global(
        self,
        name: str,
        save: bool = False,
        size_inches: Optional[tuple] = None,
        remove_title: bool = False,
        fig=None,
        ax=None,
        legend=True,
    ):
```

aeromaps/core/process.py
```diff
@@ -59,6 +64,20 @@
...
+    create_market_ask_models,
+    create_market_load_factor_models,
+    create_market_rpk_aggregator,
+    create_market_rpk_demand_model,
+    create_market_rpk_elasticity,
+    create_market_rpk_models,
+    create_market_rtk_aggregator,
+    create_market_rtk_models,
+)
+
```
ScottDelbecq, line 81:
```suggestion
# TODO: investigate if this should be handle in a model layer
# Markets registry
from aeromaps.models.air_transport.markets.market import Market
from aeromaps.models.air_transport.markets.market_manager import MarketManager
from aeromaps.models.air_transport.markets.markets_factory import (
    create_market_ask_models,
    create_market_load_factor_models,
    create_market_rpk_aggregator,
    create_market_rpk_demand_model,
    create_market_rpk_elasticity,
    create_market_rpk_models,
    create_market_rtk_aggregator,
    create_market_rtk_models,
)

```

aeromaps/models/air_transport/air_traffic/short_range_distribution.py
ScottDelbecq: This class/model seems never used.

aeromaps/models/air_transport/aircraft_fleet_and_operations/fleet/fleet_model.py
```diff
@@ -860,174 +1066,142 @@ def _get_subcategory(self, category_name: str, subcategory_name: str):
...
-                    "Warning Message - Fleet Model: long Range Aircraft: "
-                    "Average initial long-range fleet energy per ASK is higher than default energy per ASK for the old reference aircraft - "
-                    "AeroMAPS is using initial long-range fleet energy per ASK as old aircraft energy performances. "
-                    "Recent reference aircraft is introduced on first prospective year"
-                )
-                t_eis_long_range = self.parameters.prospection_start_year
-                lr_subcat.old_reference_aircraft.energy_per_ask = mean_energy_init_ask_long_range
+        if getattr(self.parameters, rpk_share_param) == 0.0:
+            mean_energy_init_ask = np.nan
+        else:
```
ScottDelbecq:
```suggestion
            warnings.warn(
                f"Warning Message - Fleet Model: {category_name} has a zero last-historical-year "
                f"RPK share — reference-aircraft calibration is skipped; the YAML-declared "
                f"reference aircraft performances are used as-is."
            )
            return
```

ScottDelbecq (changes requested):
Thanks @antSGS and team for this huge work. I made a few specific comments for which some must be addressed. 
Regarding your general comments thanks for opening issues. For the others here are some answers:
- Coupled-demand models: it is not a structuring decision for me but we will definitely in the short term end with MDAs per market
- Per market vs per field disciplines: very good point, I proposed something to trace this and to print such infos through `process.describe_models()`

aeromaps/core/multi_regional_process.py
antSGS: okay done

aeromaps/core/process.py
antSGS: done

antSGS (commented):
> This class/model seems never used. 

Deleted model, you were right

aeromaps/models/air_transport/aircraft_fleet_and_operations/fleet/fleet_model.py
antSGS: Can you explain ? Why before we had a nan and now we should just skip?

antSGS: Done

aeromaps/core/process.py
```diff
@@ -627,6 +705,191 @@ def generate_n2(self):
...
+        """
+        try:
+            for name in discipline.input_grammar.names:
+                if ":" in name:
+                    return name.split(":", 1)[0]
+        except Exception:
+            pass
+        return None
+
+    def describe_models(self, include_agnostic=False, scope=None, display=True):
```
antSGS: @ScottDelbecq nice addition. There is just something that worries me is that the categorization you've proposed is really centered on markets related models. This function might be used for a broader discovery goal, to see for instance enegry models. 

Should we change the include_agnostic to True by default? Or the intent of the function is not the one i understood?

aeromaps/models/air_transport/aircraft_fleet_and_operations/fleet/fleet_model.py
ScottDelbecq: I think this was done to skip the calibration of the reference aircraft but its cleaner to simply add a return.

aeromaps/core/process.py
ScottDelbecq: Indeed it has a broader purpose but the direct case was your concern about the market addressed. We can bump something broader in this PR or improve it later as you wish.

antSGS: Done, now broared description

ScottDelbecq (approved):

ScottDelbecq (approved):

PR #146, opened 2026-06-16
PR — Make `prospection_start_year` flexible

antSGS:
Hi everyone, i propose to submit this little PR about flexible start year. Either we can merge into main after the merge of fett-refactoring, or merge it into fleet refactoring.   
Any case I request your review here as both features are quite different.   
I'm a bit busy so i'll kepp it as a draft but testing is not fully complete. If one wants to stress test the branch, feel free to do so while updating the *reference outputs*, but i need to move on right now on a child branch to implement other features for WP1.

Important stuff: 
--> I've uploaded a test notebook in which one can strat with another year and documents how ot do it.  
--> Normally, all notebooks with default years run, but i have not chack one by one, and most importantly ** i have not updated reference outputs yet**. 
--> Claude's summary of the modifications, also documented in Flexible_start_year_plan.md with more details.   

**Source:** `feature_flex_prospection_start_year` → **Target:** `fleet-refactoring`

## Summary

Removes the hard-coded `prospection_start_year = 2020` assumption baked into the
engine through repeated `.loc[2019]` / `.loc[2020]` / `(k - 2019)` constructs and
`*_2019`-named parameters. After this change a user can set
`prospection_start_year` to any year (e.g. 2025) — supplying their own extended
`*_init` vectors and updated reference scalars — and the simulation runs
unchanged. The other bounds (`historic_start_year=2000`,
`climate_historic_start_year=1940`, `end_year=2050`) are untouched.

This is a **code-only** change: defaults stay at 2020 and no new default data is
shipped. For `prospection_start_year == 2020` the math is identical (regression
verified, see below) — the work is parameterization + renaming.

### 1. Single source of truth — `last_historical_year`
- `Parameters.last_historical_year` is a **derived `@property`** returning
  `prospection_start_year - 1`. It is computed, never stored, never persisted to
  JSON, so it cannot drift out of sync.
- `AeroMAPSModel` exposes `self.last_historical_year` to every model, so models
  read it instead of the literal `2019`.

### 2. Last-historical-year reads generalized (`.loc[2019]`, `(k - 2019)`)
All "last historic year / calibration anchor" reads now route through
`self.last_historical_year`:
- `load_factor.py`, `aircraft_efficiency.py`, `fleet_model.py`
  (reference-aircraft energy-per-ASK calibration)
- `carbon_offset.py`, `fleet_abatement_cost.py`, `co2_emissions.py`,
  `carbon_budget.py`
- air-traffic models: `rpk_market.py`, `rtk_market.py`,
  `short_range_distribution.py` (the `2019` reference-year anchor; intermediate
  waypoints kept), `price_and_income_elasticity.py`,
  `price_elasticity_logistic_income.py`

### 3. COVID-2020 patches made conditional
The unconditional `self.df.loc[2020, …] = …` COVID overrides in `load_factor.py`
and `aircraft_efficiency.py` are now guarded by `if prospection_start_year <= 2020`,
becoming no-ops when 2020 falls outside the prospective window. User-provided
historic `*_init` data is expected to already reflect the observed 2020 drop.
The `covid_*_2020` parameter **names are kept as-is** — they describe a real
historical event, not a generic pivot — and are simply ignored when the guard
short-circuits.

### 4. `*_2019` parameter rename → `*_last_historical_year`
Parameters whose `_2019` suffix encoded "last historical year" (not a 2019 event)
were renamed, e.g. `gdp_per_capita_2019`, `<mid>_rpk_share_2019`,
`<mid>_energy_share_2019`, `freight_energy_share_2019`,
`carbon_offset_baseline_level_vs_2019*`, `world_co2_emissions_2019`.

**Migration is fail-fast, not silent.** Instead of a backward-compat alias shim,
the loader raises a clear `ValueError` when it sees a stale renamed key
(`RENAMED_INPUTS` exact map + `RENAMED_INPUT_SUFFIXES` for the `_share_2019`
family), telling the user the new name. This keeps configs honest rather than
masking a half-migrated file. All in-repo configs, default `parameters.json` /
`markets.yaml`, and tutorial notebooks were migrated to the new names.

### 5. Input-length validation
`*_init` vector inputs are checked against
`prospection_start_year - historic_start_year` at load time and raise a clear
`ValueError` (offending key, expected vs actual length) instead of an obscure
pandas error.

## New artifacts
- Tutorial: **`change_the_prospection_start_year.ipynb`** with a worked
  `prospection_start_year=2025` config (`inputs_2025.json`, `config.yaml`).
- All notebooks migrated from the `2019` key to `last_historical_year`.

## Notable deviations from the original plan
- **`last_historical_year` is a derived property** rather than an attribute set in
  `_initialize_inputs` — cannot drift, never persisted.
- **No backward-compat alias shim.** The plan proposed silent aliases with a
  deprecation warning; the implementation uses a hard rename guard that errors,
  so stale configs fail loudly with the corrected name.

## Verification
- **Default regression (`01_basic`):** zero diff in `outputs.json`
  (`prospection_start_year == 2020` math unchanged).
- **`pytest aeromaps/`:** model-level regressions covered.
- **Custom 2025 scenario:** loads (Phase E gate passes), `load_factor.loc[2020]`
  comes from the user's historic `rpk_init/ask_init` (not `covid_load_factor_2020`),
  offset baseline reads from `last_historical_year`.

## Out of scope
- Climate historical CSV extension (`temperature_historical_dataset.csv`) — the
  user supplies an extended CSV for `prospection_start_year > 2020` with climate
  models enabled.
- Publication notebooks keep their hard-coded year arrays unless requested.


antSGS:
Chained PRs 3/4

PR #157, opened 2026-08-20
fix(mda): unpin the residual floor, stop column duplication, allow global disciplines in unified_mda

antSGS:
Three independent fixes to the MDA / multi-regional machinery, found while spiking a
global fuel-market discipline. The spike code itself is **not** in this PR — only the
general fixes, one per commit, each reviewable on its own.

Two of them are latent bugs that affect any AeroMAPS user today; the third is the small
capability that was missing from `unified_mda`.

---

## 1. Disciplines mutating their MDA inputs — the residual floor (`33d0f31d`)

**The bug.** GEMSEO snapshots the previous MDA iterate with `self.io.data.copy()`, which
is a **shallow** copy: the `pd.Series` it holds are the very objects handed to each
discipline's `compute()` as `input_data`. A discipline that mutates one in place
therefore rewrites the snapshot the solver is about to difference against, and the
residual of that coupling variable stops meaning anything.

`CO2Emissions` did exactly that:

```python
co2_emission_factor = input_data[f"{energy_type}_mean_co2_emission_factor"]
co2_emission_factor.fillna(0, inplace=True)   # comment said "locally fill" — inplace is not local
```

For an all-NaN emission factor (a scenario with no hydrogen and no electric aircraft, i.e.
most scenarios) the **output** side of the residual converts to the `-999999` NaN sentinel
while the **input** side has already been rewritten to `0`. The difference is a constant
`999999` per element, forever. That pins the normalized residual at ~1.6e-6 — above the
`tolerance=1e-10` that `create_process` asks for.

Confirmed by object identity: same `id()`, `nan=51` at the top of `_iterate_once`, `nan=0`
by the time the residual is computed.

`BottomUpCapacity` had the same defect in a worse form — it appended "virtual years" to
`{pathway}_energy_consumption` and re-sorted it in place, changing a coupling variable's
**length**.

**Why it matters.** With stock settings (`tolerance=1e-10`, `max_mda_iter=200`), on
`aeromaps/tests/tested_configs/config_elasticity_demand.yaml`:

| | iterations | final residual | converged? |
|---|---|---|---|
| `main` | 160 (gave up: 8 consecutive unsuccessful) | 3.34e-07 | **no** |
| this PR | 109 | 6.19e-11 | yes |

So on `main` the price-elastic demand loop silently never reaches the tolerance the code
deliberately asks for. Note the failure is silent in *both* directions — a discipline that
mutated an input toward its output would drive a residual artificially to zero and report
a convergence that never happened.

**What is *not* changed.** The `-999999` sentinel in `CustomDataConverter` was never the
problem and is untouched. Once nothing mutates the snapshot it works exactly as designed:
NaN maps to the sentinel on both sides, the difference is zero, and the variable
contributes nothing to the residual. **No NaN is turned into a fake zero** — all-NaN
columns stay all-NaN (214 of them in the two-region scenario I checked).

**The fix**: non-mutating `.fillna(0)` in one model, `.copy()` before extending in the
other. Plus `aeromaps/tests/core/test_mda_input_mutation.py`, which asserts the invariant
directly across every discipline of a chain and names the offending discipline and
variable when it fails. Verified in both directions — it does fail when either mutation
is reintroduced.

> **Reviewer note:** this is the invariant worth taking away from the PR — *never mutate a
> value you receive in `input_data`*. `.fillna(0)` not `.fillna(0, inplace=True)`,
> `.copy()` before any `.loc[...] =` or `sort_index(inplace=True)`.

---

## 2. Every output column duplicated on repeated `compute()` (`42a1af2f`)

**The bug.** `_update_data_from_unified_mda` and `_aggregate_regional_outputs` both built
`data["vector_outputs"]` / `data["climate_outputs"]` by concatenating the new series onto
the *existing* frame. A second `compute()` on the same process appended a second copy of
every column instead of refreshing it. Once a column name is duplicated, `df[col]` returns
a **DataFrame instead of a Series** and downstream code misbehaves in ways that are hard
to trace back here.

Reproducible on the stock two-region tutorial, unchanged from `main`, **both execution
modes**:

```
main      cols per compute() = [823, 1646, 2469]   duplicated = 1646
this PR   cols per compute() = [823,  823,  823]   duplicated = 0
```

**The fix.** Both call sites go through a small `_concat_series` helper that drops the
names being rewritten before concatenating — still a single `concat`, so the
anti-fragmentation intent of the original code is preserved. Values are bit-identical run
to run.

---

## 3. `unified_mda` could not host a non-namespaced discipline (`af19caf0`)

This one is a **capability addition rather than a bug fix**, so review it as such — it is
the largest diff and the only one that adds config surface. It is also independently
droppable if you would rather take 1 and 2 alone.

**The gap.** `unified_mda` had no way to host a discipline that reads one region's
variable and writes another's. The only route into the chain was `top_level_models`, which
namespaces the discipline to `{global_namespace}:` and thereby confines it to the
aggregated series. `RegionalAggregator` was the sole non-namespaced discipline and it was
hard-wired.

**The addition.** A `regionalisation.global_models` block, with the same
standards/customs structure as a regional models block. A global model is wrapped
**without** namespacing, so its grammar is written directly in namespaced terms
(`{region}:var`), exactly like the aggregator's. Unlike `_wrap_top_level_model`,
`_wrap_global_model` injects the region list and the global namespace and *then* calls
`custom_setup()`, so a model that builds its I/O dynamically from the region list gets a
real grammar instead of the placeholder from its `__init__`.

Declaring a global model under `separate_processes` now raises instead of being silently
ignored — that mode cannot solve a loop across regions.

Also in this commit:

- **Harvesting of global-model outputs** in `_update_data_from_unified_mda`. Previously
  only `{global_namespace}:`-prefixed keys were collected from `local_data`, so a global
  model's outputs never reached `data["vector_outputs"]`.
- `_register_models_into` factored out of `_register_top_level_models`.

~~A `regionalisation.mda` block to override the hard-coded MDAChain settings.~~ Removed in
`83d203eb` after review — MDA settings stay hard-coded and are tuned from the notebook on
the GEMSEO objects, as `process.py` does. The `MDAChain` construction is byte-identical to
`main` again, apart from a comment recording the tuning trap: `over_relaxation_factor` and
`acceleration_method` are not `BaseMDASettings` fields, so `MDAChain` never forwards them
to its inner MDAs — they must be set on `mda_chain.inner_mdas[i]`.

**Defaults are unchanged** — a config without a `global_models` or `mda` block behaves
exactly as before.

---

## Verification

- `pytest aeromaps/tests/core` — **28 passed** on this branch (serial run; running two
  pytest processes concurrently makes them collide over `tested_configs/tmp`).
- Two-region tutorial, both modes, on this branch: identical results
  (`co2_emissions` 2050 = 2.0801152004054027 in `unified_mda` and in
  `separate_processes`), stable column count across three consecutive `compute()` calls.
- Before/after numbers in sections 1 and 2 measured by running the same script against a
  clean `main` worktree and against this branch.

**Coverage gap worth flagging:** `aeromaps/core/multi_regional_process.py` has *no*
automated test coverage on `main` — nothing under `aeromaps/tests/` exercises
`MultiRegionalProcess` at all. Commits 2 and 3 are therefore verified by the manual runs
above, not by CI. Adding a multi-regional test is out of scope here but is the obvious
follow-up.

## Review order suggestion

`33d0f31d` first (smallest, highest impact), then `42a1af2f`, then `af19caf0`.

### DIFF
--- a/aeromaps/core/multi_regional_process.py
+++ b/aeromaps/core/multi_regional_process.py
@@ -49,6 +49,34 @@ from aeromaps.plots.single_scenario import (
 ExecutionMode = Literal["unified_mda", "separate_processes"]
 
 
+def _concat_series(frame: pd.DataFrame, series: Dict[str, pd.Series]) -> pd.DataFrame:
+    """Add ``series`` as columns of ``frame``, replacing any column of the same name.
+
+    Columns are concatenated in one pass to avoid DataFrame fragmentation. Names
+    already present in ``frame`` are dropped first: ``compute()`` may be called more
+    than once on the same process, and without this the second call appended a second
+    copy of every column instead of refreshing it.
+
+    Parameters
+    ----------
+    frame
+        The existing DataFrame (returned unchanged when ``series`` is empty).
+    series
+        Mapping of column name to Series.
+
+    Returns
+    -------
+    pd.DataFrame
+        The frame with the new columns.
+    """
+    if not series:
+        return frame
+    stale = [col for col in frame.columns if col in series]
+    if stale:
+        frame = frame.drop(columns=stale)
+    return pd.concat([frame] + [pd.DataFrame({k: v}) for k, v in series.items()], axis=1)
+
+
 class _GlobalOutputsView:
     """Single-scenario-process facade over a MultiRegionalProcess' aggregated outputs.
 
@@ -983,18 +1011,8 @@ class MultiRegionalProcess(AeroMAPSProcess):
                 self.data["float_outputs"][key] = float(value)
 
         # Build DataFrames efficiently using concat to avoid fragmentation
-        if vector_series:
-            self.data["vector_outputs"] = pd.concat(
-                [self.data["vector_outputs"]]
-                + [pd.DataFrame({k: v}) for k, v in vector_series.items()],
-                axis=1,
-            )
-        if climate_series:
-            self.data["climate_outputs"] = pd.concat(
-                [self.data["climate_outputs"]]
-                + [pd.DataFrame({k: v}) for k, v in climate_series.items()],
-                axis=1,
-            )
+        self.data["vector_outputs"] = _concat_series(self.data["vector_outputs"], vector_series)
+        self.data["climate_outputs"] = _concat_series(self.data["climate_outputs"], climate_series)
 
     def _update_data_from_unified_mda(self):
         """Update all data structures from unified MDA results.
@@ -1053,18 +1071,8 @@ class MultiRegionalProcess(AeroMAPSProcess):
                     self.data["float_outputs"][key] = float(value)
 
         # Build DataFrames efficiently using concat to avoid fragmentation
-        if vector_series:
-            self.data["vector_outputs"] = pd.concat(
-                [self.data["vector_outputs"]]
-                + [pd.DataFrame({k: v}) for k, v in vector_series.items()],
-                axis=1,
-            )
-        if climate_series:
-            self.data["climate_outputs"] = pd.concat(
-                [self.data["climate_outputs"]]
-                + [pd.DataFrame({k: v}) for k, v in climate_series.items()],
-                axis=1,
-            )
+        self.data["vector_outputs"] = _concat_series(self.data["vector_outputs"], vector_series)
+        self.data["climate_outputs"] = _concat_series(self.data["climate_outputs"], climate_series)
 
     # =========================================================================
     # Data Access Methods
--- a/aeromaps/core/multi_regional_process.py
+++ b/aeromaps/core/multi_regional_process.py
@@ -25,6 +25,7 @@ from typing import Dict, List, Optional, Literal
 # Third-party imports
 import numpy as np
 import pandas as pd
+from gemseo.algos.sequence_transformer.acceleration import AccelerationMethod
 from gemseo.mda.mda_chain import MDAChain
 from tqdm.auto import tqdm
 
@@ -685,22 +686,37 @@ class MultiRegionalProcess(AeroMAPSProcess):
 
         # Build MDAChain
         # TODO: Make these kwargs available at a higher level (e.g. config file).
-        #
         # Until then they are tuned from the notebook on the GEMSEO objects themselves,
-        # e.g. ``process.mda_chain.settings.tolerance = 1e-10``. One trap is worth
-        # knowing about: ``MDAChain`` forwards to its inner MDAs only the settings that
-        # belong to ``BaseMDASettings`` (``tolerance``, ``max_mda_iter``, ``warm_start``,
-        # ``log_convergence``, ...). ``over_relaxation_factor`` and
-        # ``acceleration_method`` are NOT in that set, so setting them here or on the
-        # chain configures the outer chain only and is silently ignored by the solver
-        # that actually iterates -- they have to be set on ``mda_chain.inner_mdas[i]``
-        # (or passed as ``inner_mda_settings``).
+        # e.g. ``process.mda_chain.settings.tolerance = 1e-10``.
+        #
+        # ``inner_mda_settings`` is listed separately on purpose: it is the only route to
+        # the relaxation/acceleration knobs. ``MDAChain`` forwards to its inner MDAs only
+        # the settings belonging to ``BaseMDASettings`` (``tolerance``, ``max_mda_iter``,
+        # ``warm_start``, ``log_convergence``, ...), and ``over_relaxation_factor`` /
+        # ``acceleration_method`` are NOT in that set, so passing them as ``MDAChain``
+        # kwargs configures the outer chain alone and is silently ignored by the solver
+        # that actually iterates.
+        #
+        # The values below are the MDAGaussSeidel defaults (no damping, no acceleration),
+        # so the chain behaves exactly as before. To change them from a notebook, assign
+        # to the *properties* of the inner solver, not to its settings::
+        #
+        #     process.mda_chain.inner_mdas[0].over_relaxation_factor = 0.7
+        #     process.mda_chain.inner_mdas[0].acceleration_method = "Alternate2Delta"
+        #
+        # ``BaseMDASolver.__init__`` builds its ``RelaxationAcceleration`` from the
+        # settings once and never re-reads them; those two properties write through to it,
+        # whereas assigning to ``inner_mdas[0].settings.*`` afterwards has no effect.
         self.mda_chain = MDAChain(
             disciplines=all_disciplines,
             tolerance=1e-5,
             initialize_defaults=True,
             inner_mda_name="MDAGaussSeidel",
             log_convergence=True,
+            inner_mda_settings={
+                "over_relaxation_factor": 1.0,
+                "acceleration_method": AccelerationMethod.NONE,
+            },
         )
 
         logging.info(f"Unified MDAChain created with {len(all_disciplines)} disciplines")
--- a/aeromaps/core/multi_regional_process.py
+++ b/aeromaps/core/multi_regional_process.py
@@ -683,47 +683,27 @@ class MultiRegionalProcess(AeroMAPSProcess):
 
         self.disciplines = all_disciplines
 
-        self.mda_chain = MDAChain(disciplines=all_disciplines, **self._mda_settings())
+        # Build MDAChain
+        # TODO: Make these kwargs available at a higher level (e.g. config file).
+        #
+        # Until then they are tuned from the notebook on the GEMSEO objects themselves,
+        # e.g. ``process.mda_chain.settings.tolerance = 1e-10``. One trap is worth
+        # knowing about: ``MDAChain`` forwards to its inner MDAs only the settings that
+        # belong to ``BaseMDASettings`` (``tolerance``, ``max_mda_iter``, ``warm_start``,
+        # ``log_convergence``, ...). ``over_relaxation_factor`` and
+        # ``acceleration_method`` are NOT in that set, so setting them here or on the
+        # chain configures the outer chain only and is silently ignored by the solver
+        # that actually iterates -- they have to be set on ``mda_chain.inner_mdas[i]``
+        # (or passed as ``inner_mda_settings``).
+        self.mda_chain = MDAChain(
+            disciplines=all_disciplines,
+            tolerance=1e-5,
+            initialize_defaults=True,
+            inner_mda_name="MDAGaussSeidel",
+            log_convergence=True,
+        )
 
         logging.info(f"Unified MDAChain created with {len(all_disciplines)} disciplines")
-
-    def _mda_settings(self) -> dict:
-        """Resolve the MDAChain settings, overridable from the configuration file.
-
-        Read from an optional ``regionalisation.mda`` block::
-
-            regionalisation:
-              mda:
-                tolerance: 1.0e-10
-                max_mda_iter: 200
-                inner_mda_name: MDAGaussSeidel
-                log_convergence: true
-                inner_mda_settings:
-                  over_relaxation_factor: 0.7
-                  acceleration_method: Alternate2Delta
-
-        ``inner_mda_settings`` is load-bearing and easy to get wrong: ``MDAChain``
-        forwards to its inner MDAs only the settings that belong to
-        ``BaseMDASettings`` (``tolerance``, ``max_mda_iter``, ``warm_start``,
-        ``log_convergence``, ...). ``over_relaxation_factor`` and
-        ``acceleration_method`` are NOT in that set, so passing them as top-level
-        ``MDAChain`` kwargs configures the outer chain only and is silently ignored by
-        the solver that actually iterates. They must go through ``inner_mda_settings``.
-
-        Returns
-        -------
-        dict
-            Keyword arguments for the ``MDAChain`` constructor.
-        """
-        settings = {
-            "tolerance": 1e-5,
-            "initialize_defaults": True,
-            "inner_mda_name": "MDAGaussSeidel",
-            "log_convergence": True,
-        }
-        user_settings = self._regionalisation_config.get("mda", {}) or {}
-        settings.update(user_settings)
-        return settings
 
     def _initialize_data_containers(self):
         """Initialize data containers following AeroMAPSProcess structure.
--- a/aeromaps/core/multi_regional_process.py
+++ b/aeromaps/core/multi_regional_process.py
@@ -191,6 +191,10 @@ class MultiRegionalProcess(AeroMAPSProcess):
         # Load optional top-level models that run on aggregated (global-namespace) data
         self._load_top_level_models()
 
+        # Load optional global models: disciplines that are NOT namespaced and whose
+        # grammar spans several regions at once (see _load_global_models).
+        self._load_global_models()
+
         # Reference parameters (year indexing, etc.) from the first region. Needed
         # before building top-level disciplines so models can initialize their dataframes.
         self.parameters = self._regional_processes[self._region_ids[0]].parameters
@@ -369,17 +373,29 @@ class MultiRegionalProcess(AeroMAPSProcess):
         models
             Dict whose values are either AeroMAPSModel instances or nested dicts of them.
         """
+        self._register_models_into(models, self._top_level_model_names)
+
+    def _register_models_into(self, models, name_list):
+        """Recursively register AeroMAPSModel instances into ``self.models``.
+
+        Parameters
+        ----------
+        models
+            Dict whose values are either AeroMAPSModel instances or nested dicts of them.
+        name_list
+            List that collects the registered model names (mutated in place).
+        """
         from aeromaps.models.base import AeroMAPSModel
 
         for key, value in models.items():
             if isinstance(value, dict):
-                self._register_top_level_models(value)
+                self._register_models_into(value, name_list)
             elif isinstance(value, AeroMAPSModel):
                 self.models[value.name] = value
-                self._top_level_model_names.append(value.name)
+                name_list.append(value.name)
             else:
                 raise TypeError(
-                    f"Top-level model entry '{key}' is not an AeroMAPSModel instance "
+                    f"Model entry '{key}' is not an AeroMAPSModel instance "
                     f"(got {type(value).__name__})."
                 )
 
@@ -455,6 +471,129 @@ class MultiRegionalProcess(AeroMAPSProcess):
             AeroMAPSCustomModelWrapper(model=self.models["aggregator"])
         ] + self._build_namespaced_top_level_disciplines()
 
+    # =========================================================================
+    # Global (non-namespaced) models
+    # =========================================================================
+
+    def _load_global_models(self):
+        """Load optional *global* models: disciplines deliberately kept OUT of namespacing.
+
+        A global model differs from a top-level model on one axis only, but it is a
+        decisive one:
+
+        - a **top-level** model is namespaced to ``self._global_namespace``, so it can
+          only see the aggregated ``{global_namespace}:*`` series. It is a consumer of
+          aggregates, downstream of the aggregator.
+        - a **global** model is *not* namespaced at all. Its grammar is written directly
+          in namespaced terms (``{region}:var``), exactly like ``RegionalAggregator``'s,
+          so it can read one region's variable and write another's. That is what makes
+          it able to close an inter-regional coupling loop — e.g. a fuel market whose
+          clearing price in every region depends on total demand across all regions.
+
+        Declared under ``regionalisation.global_models`` with the same
+        ``standards``/``customs`` structure as a regional ``models`` block::
+
+            regionalisation:
+              execution_mode: unified_mda
+              global_models:
+                customs:
+                  fuel_market: "models/fuel_market.py::FuelMarket"
+
+        Because the region list is not knowable by the model itself, it is injected
+        before ``custom_setup()`` runs (see :meth:`_wrap_global_model`), mirroring how
+        ``AeroMAPSProcess`` injects ``markets`` / ``pathways_manager`` into regional
+        models before calling their ``custom_setup()``.
+
+        Global models only make sense in ``unified_mda`` mode: ``separate_processes``
+        solves each region in isolation and cannot close a loop across regions.
+        """
+        self._global_model_names = []
+
+        global_config = self._regionalisation_config.get("global_models", {})
+        if not global_config:
+            return
+
+        loaded = {}
+        for model_name in global_config.get("standards", []):
+            if hasattr(aeromaps_models, model_name):
+                loaded[model_name] = getattr(aeromaps_models, model_name)
+            else:
+                raise ValueError(
+                    f"Global model '{model_name}' specified in 'regionalisation."
+                    f"global_models.standards' is not found in aeromaps.core.models."
+                )
+
+        customs = global_config.get("customs", None)
+        if customs is not None:
+            loaded.update(self._load_custom_models_from_config(customs))
+
+        self._register_models_into(loaded, self._global_model_names)
+
+        logging.info(
+            f"Loaded {len(self._global_model_names)} global model(s): "
+            f"{self._global_model_names}"
+        )
+
+    def _wrap_global_model(self, model):
+        """Initialize and wrap a single global model as a NON-namespaced discipline.
+
+        Unlike :meth:`_wrap_top_level_model` this method:
+
+        - injects the region list (``model.regions``) and the global namespace
+          (``model.global_namespace``) when the model declares those attributes, and
+        - calls ``custom_setup()`` afterwards, so a model that builds its I/O names
+          dynamically from the region list gets a real grammar instead of the
+          placeholder set in its ``__init__``.
+
+        ``_initialize_df()`` is called twice on purpose: once before ``custom_setup()``
+        so the model has its year index while building its grammar, and once after so
+        that ``_coupling_defaults`` — which typically needs the region list — is rebuilt
+        against the final configuration.
+
+        Parameters
+        ----------
+        model
+            The AeroMAPSModel instance to wrap.
+
+        Returns
+        -------
+        Discipline
+            The wrapped, NOT namespaced, GEMSEO discipline.
+        """
+        model.parameters = self.parameters
+        model._initialize_df()
+
+        if hasattr(model, "climate_historical_data"):
+            raise NotImplementedError(
+                f"Global model '{model.name}' requires climate_historical_data, which is "
+                "not available at the global level."
+            )
+
+        if hasattr(model, "regions"):
+            model.regions = list(self._region_ids)
+        if hasattr(model, "global_namespace"):
+            model.global_namespace = self._global_namespace
+
+        if hasattr(model, "custom_setup"):
+            model.custom_setup()
+
+        # Rebuild the year frames / coupling seeds now that the model is configured.
+        model._initialize_df()
+
+        if getattr(model, "model_type") == "custom":
+            return AeroMAPSCustomModelWrapper(model=model)
+        return AeroMAPSAutoModelWrapper(model=model)
+
+    def _build_global_disciplines(self):
+        """Build the global model disciplines, deliberately WITHOUT namespacing.
+
+        Returns
+        -------
+        list
+            List of GEMSEO disciplines (empty if no global models).
+        """
+        return [self._wrap_global_model(self.models[name]) for name in self._global_model_names]
+
     def _setup_separate_processes(self):
         """Set up separate_processes mode.
 
@@ -464,6 +603,14 @@ class MultiRegionalProcess(AeroMAPSProcess):
         processes, fed with their namespaced outputs. This replaces the previous bespoke
         aggregator call with a normal MDA execution.
         """
+        if self._global_model_names:
+            raise NotImplementedError(
+                f"Global model(s) {self._global_model_names} are declared under "
+                "'regionalisation.global_models', but execution_mode is "
+                "'separate_processes'. A global model exists to close a coupling loop "
+                "across regions, which only 'unified_mda' can solve."
+            )
+
         self.mda_chain = None
         self.disciplines = self._build_top_level_disciplines()
         self._top_level_mda_chain = MDAChain(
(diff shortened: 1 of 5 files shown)


aeromaps/core/multi_regional_process.py
```diff
@@ -455,6 +499,129 @@ def _build_top_level_disciplines(self):
...
+        -------
+        Discipline
+            The wrapped, NOT namespaced, GEMSEO discipline.
+        """
+        model.parameters = self.parameters
+        model._initialize_df()
+
+        if hasattr(model, "climate_historical_data"):
+            raise NotImplementedError(
+                f"Global model '{model.name}' requires climate_historical_data, which is "
```
antSGS, line 603: the regional aggregator should provide it no? Is that feature not available yet?

aeromaps/core/multi_regional_process.py
```diff
@@ -464,6 +631,14 @@ def _setup_separate_processes(self):
         processes, fed with their namespaced outputs. This replaces the previous bespoke
         aggregator call with a normal MDA execution.
         """
+        if self._global_model_names:
+            raise NotImplementedError(
+                f"Global model(s) {self._global_model_names} are declared under "
+                "'regionalisation.global_models', but execution_mode is "
+                "'separate_processes'. A global model exists to close a coupling loop "
+                "across regions, which only 'unified_mda' can solve."
+            )
```
antSGS, line 647: What would be neceessary to make the separate processes mode working? I mean processes could be seen as individual disciplines of the gloabl process right?

aeromaps/core/multi_regional_process.py
```diff
@@ -502,20 +677,54 @@ def _setup_unified_mda(self):
...
+        self.mda_chain = MDAChain(disciplines=all_disciplines, **self._mda_settings())
 
         logging.info(f"Unified MDAChain created with {len(all_disciplines)} disciplines")
 
+    def _mda_settings(self) -> dict:
+        """Resolve the MDAChain settings, overridable from the configuration file.
+
+        Read from an optional ``regionalisation.mda`` block::
+
+            regionalisation:
```
antSGS: I am not yet comfortable with teh idea of having MDA parameters specified in the config file directly. That a long term nice to have but as you can see in process.py the mda config are done through gemseo attributes defined in teh notebook. I think we should stick to that for now.

antSGS: **Claude answer**

Agreed, reverted in 83d203eb.

`_mda_settings()` and the `regionalisation.mda` block are gone. The `MDAChain`
construction in `_setup_unified_mda` is now byte-identical to `main` again (same
`tolerance=1e-5`, same `inner_mda_name`), plus one comment, and the original
`TODO: Make these kwargs available at a higher level` is back where it was.

I left the comment because it records the one thing that bit me while tuning the chain
from a notebook, which is exactly the workflow you're asking to keep:

> `MDAChain` forwards to its inner MDAs only the settings that belong to
> `BaseMDASettings` (`tolerance`, `max_mda_iter`, `warm_start`, `log_convergence`, ...).
> `over_relaxation_factor` and `acceleration_method` are **not** in that set, so setting
> them on the chain configures the outer chain only and is silently ignored by the solver
> that actually iterates — they have to be set on `mda_chain.inner_mdas[i]`.

That one cost me an afternoon and a wrong conclusion ("damping does nothing here"), so
it seemed worth writing down next to the constructor. Happy to drop the comment too if
you'd rather keep the diff at zero.

One observation while I was in there, entirely for you to judge — **not** something I've
changed: `_setup_unified_mda` uses `tolerance=1e-5`, while `AeroMAPSProcess` uses `1e-10`
with an explicit comment that `1e-5` is too loose to resolve the
`doc_net_energy_per_rpk_mean` ↔ `rpk` loop ("at 1e-5 the solver reports convergence while
that coupling is still ~25% off in SAF-type scenarios"). So a multi-regional run is
currently held to a tolerance the single-region code says is not enough. It was
unreachable anyway before the residual fix in this PR; now it would be. Worth a separate
look, but it changes results so it doesn't belong in this PR.

aeromaps/core/multi_regional_process.py
antSGS: **Claude answer**

Not available yet, and this particular line isn't new — it's copied verbatim from the
guard that's already on `main` in `_wrap_top_level_model`:

```python
unsupported = []
if hasattr(model, "climate_historical_data"):
    unsupported.append("climate_historical_data")
```

I mirrored it so a global model fails the same way a top-level model already does, rather
than failing later with an `AttributeError`.

On the aggregator: it does aggregate climate **outputs** — `RegionalAggregator` detects
climate-length series and routes them to `df_climate`, so `overall:temperature_increase`
and friends do get built. But `climate_historical_data` is a different animal. It's an
*input* array, not an output: `AeroMAPSProcess._initialize_climate_historical_data()`
builds it from the partitioning JSON (or a CSV) as
`np.column_stack([years, co2_emissions, nox_emissions, h2o_emissions, soot_emissions, sulfur_emissions, distance])`
— observed history, loaded per regional process from that region's partitioning file. The
aggregator never sees it because it isn't a discipline output, and
`MultiRegionalProcess` has no `_initialize_climate_historical_data` of its own.

**What it would take.** Mechanically, very little: those six columns are all extensive, so
summing the per-region arrays column-wise (keeping the years column) reconstructs a global
one, and it would slot straight into both `_wrap_top_level_model` and
`_wrap_global_model`.

**Why I'd want your call before doing it.** That sum is only the *world* history if the
regions partition the world exhaustively and disjointly. That's the assumption behind the
partitioning files, but nothing enforces it for an arbitrary region set — the two-region
Europe tutorial in this repo covers Europe only, and summing its two partitions gives a
European history, not a global one. A climate model fed a partial-world emissions history
returns a temperature that looks authoritative and is meaningless. So the honest version
of the feature needs either a completeness check on the region set or an explicit
"these regions are the whole world" flag in the config.

Happy to do it as a follow-up PR if you want it — it's small, but it's a modelling
decision more than a plumbing one, and it's orthogonal to the three fixes here.

aeromaps/core/multi_regional_process.py
antSGS: **Claude answer**

Your instinct is right, and it's the neat version too: in GEMSEO an `MDAChain` *is* a
`Discipline`, so each regional process's own `mda_chain` can be namespaced and dropped
into an outer MDA alongside the aggregator and the global models. Roughly:

```
outer MDA
├── apply_namespace_to_discipline(process["EU_DOM"].mda_chain, "EU_DOM")
├── apply_namespace_to_discipline(process["EU_INT"].mda_chain, "EU_INT")
├── RegionalAggregator
└── FuelMarket            # non-namespaced, reads {region}:x, writes {region}:y
```

So it's not a large piece of work — mostly namespacing whole chains instead of individual
disciplines, and replacing the `for region_id, regional_process in ...: regional_process.compute()`
loop in `_compute_separate_processes` with an outer `MDAChain.execute()`.

**But I don't think it's worth building, and that's why I made it raise rather than
silently no-op.** Two reasons:

1. **It converges to the same answer as `unified_mda`, more slowly.** Every outer
   iteration re-solves each region's inner MDA *to convergence* just to produce one
   updated coupling value. `unified_mda` sweeps at discipline granularity inside a single
   SCC and gets there in far fewer discipline executions. Nested MDA pays off when the
   subproblems are expensive and nearly decoupled; here the inter-regional coupling is
   exactly what's tight.

2. **It destroys the thing `separate_processes` is for.** That mode exists because regions
   are independent, which is what makes `_compute_separate_processes(parallel=True)`
   valid. The moment a global discipline closes a loop across regions, they aren't
   independent any more. You could keep some parallelism with an outer `MDAJacobi`
   (all regions evaluated from the same previous iterate, so still fan-out-able), but
   Jacobi converges more slowly per iteration than Gauss-Seidel and I couldn't get it to
   run on the real chain at all during the spike — it dies with
   `ValueError: array must not contain infs or NaNs`.

So my read is that `unified_mda` is the right and only home for cross-regional coupling,
and `separate_processes` should stay the fast embarrassingly-parallel path for scenarios
that genuinely have no such coupling. Hence the explicit `NotImplementedError` naming the
offending models, instead of a config that appears to work and quietly gives you an
unconverged answer.

If you disagree and want the nested mode anyway, say so and I'll scope it properly — I'd
just rather that be a deliberate decision than a side effect of this PR.

aeromaps/core/multi_regional_process.py
antSGS: **Claude answer**

Done in cb2e709b — `inner_mda_settings` is now passed explicitly on the unified
`MDAChain`, with the `MDAGaussSeidel` defaults, so behaviour is unchanged and the knobs
have a visible home next to the ones that were already there:

```python
self.mda_chain = MDAChain(
    disciplines=all_disciplines,
    tolerance=1e-5,
    initialize_defaults=True,
    inner_mda_name="MDAGaussSeidel",
    log_convergence=True,
    inner_mda_settings={
        "over_relaxation_factor": 1.0,
        "acceleration_method": AccelerationMethod.NONE,
    },
)
```

I checked all four claims in the comment against GEMSEO 6.2.0 on a minimal coupled chain
rather than trusting the docs:

```
1. via inner_mda_settings                -> orf=0.7  accel=Alternate2Delta
2. via MDAChain kwargs                   -> orf=1.0  accel=NoTransformation  (ignored)
3a. after .settings.orf = 0.5            -> effective orf=1.0                (no effect)
3b. after .over_relaxation_factor = 0.5  -> effective orf=0.5
4. MDAGaussSeidel defaults               -> orf=1.0  accel=NoTransformation
```

Line 3 is the part I'd flag for the notebook workflow: `BaseMDASolver.__init__` builds its
`RelaxationAcceleration` from the settings **once** and never re-reads them, so from a
notebook you have to assign to the properties, not to `.settings`:

```python
process.mda_chain.inner_mdas[0].over_relaxation_factor = 0.7
process.mda_chain.inner_mdas[0].acceleration_method = "Alternate2Delta"
```

One thing worth knowing, and probably the reason nobody has hit any of this before: on the
stock two-region tutorial `len(process.mda_chain.inner_mdas) == 0`. That scenario has no
coupling loop at all, so no inner MDA is ever created and these settings have nothing to
act on. They start mattering as soon as a discipline closes a cycle across regions —
which is exactly what the `global_models` hook in this PR makes possible.

Two-region tutorial re-run after the change: identical in both modes
(`co2_emissions` 2050 = 2.0801152004054027), column count stable across repeated
`compute()`.

ScottDelbecq (approved):
Nicely spotted and corrected!
Thanks!

iancostalves:
A few general comments for each problem raised:

1. Good point the convergence, some of my WCTR applications had convergence issues that were also silent, I think that overall we can say that we need a more effective way of logging things (too many small warnings currently, and too few warnings over important stuff)
2. A bit difficult to understand this issue without some code to reproduce it, could you add maybe one file that demonstrates each failure mode? (even if they'll be deleted later)
3. Big picture stuff there, as all models can read and mutate the io dataframe this was bound to happen eventually. In my opinion, it makes sense to deal with this issue either at the same time or before we do porting to JAX (i think we'll need to get rid of dfs for computing mdas/opts and just fill the df after all was iterated in numpy/jaxarrays)
4. Fixed on gemseo's v6.1.0 `SequenceTransformer class now has a set_bounds method` https://gemseo.readthedocs.io/en/stable/software/changelog.html#version-6-1-0-2025-03-17
5. Claude already has some elements of answer there, maybe instead of computing aggregated DOC/ASK with (volumeA *qA + volumeB * qB)/(volA + volB) which yields the 0/0 you could simply make the calculation based on shares (shareA * qA + shareB * qB) even if volA+volB is zero the shares are still well defined because that's the base of computations in both fleet and energy
6. OK with Claude's solution, would have to be rethinked depending on what is done on 3
7. not much to say here
8. Good catch, I wonder how the pass-through is going to be modeled though... in the WCTR applications a delay is applied between airline cost and consumer-felt cost that could be used to compute a dynamic pass-through instead of a static one
9. Also fixed on gemseo's v6.1.0 `compare_dict_of_arrays function now handles NaNs`
10. (Verification): the coverage gap in the multi-region might be good to address if we are to capitalize them on the app. One thing I noticed in my notebooks is that since the generic markets PR #145 the process is now slower to compute (cumbersome especially for MDAs), might be good to re-make the benchmark from a_r_m and compare with what we had before this PR

iancostalves:
I have iterated some of the points I mentioned with Claude. This is more of a proposition from my side rather than an imposition. Still open for review and discussion. Here is Claude's resume:

> Follow-up on your review. Below: what each of your ten points ended up as, then the
> detail on the three that needed real work — the NaN sentinel, the `gemseo` pin, and
> pass-through.
> 
> All numbers re-measured on this branch.
> 
> ## Where each of your points landed
> 
> | # | your point | outcome |
> |---|---|---|
> | 1 | logging: too many small warnings, too few important ones | **partly done.** The convergence message now separates the three causes and names the coupling variables holding the residual up, and `RPKElasticity`'s bound warning is gone entirely (§2 below). The wider clean-up is an issue. |
> | 2 | hard to follow without code to reproduce each failure mode | **done.** `aeromaps/notebooks/dev/mda_failure_modes.py` puts each defect back at runtime on a shipped scenario and prints before/after. Not collected by pytest; delete it whenever it stops earning its place. |
> | 3 | models can mutate the io dataframe; deal with it at/before the JAX port | **agreed, deferred.** A test now makes any violation loud (`test_mda_input_mutation.py`, plus `test_mda_residual_floor.py` showing what a violation costs). The structural fix is arrays-in/arrays-out; issue raised, and it is a prerequisite for JAX rather than a side-effect of it. |
> | 4 | `set_bounds` exists since gemseo 6.1.0 | **done, and it now works.** It could not be used while the NaN sentinel existed; removing the sentinel unblocked it, and the in-model airfare clip is gone. §2. |
> | 5 | compute the aggregate DOC/ASK from shares, not volumes | **done.** `ASKAggregator` publishes `ask_<mid>_share`; the six DOC means weight by it. Bit-identical wherever there is traffic. |
> | 6 | duplicated columns — depends on what happens with (3) | **done as merged**, plus the first automated coverage `MultiRegionalProcess` has ever had. |
> | 7 | nothing to say | — |
> | 8 | how should pass-through be modelled? a delay could make it dynamic | **answered, not implemented.** §3 — including why it is hardcoded at 1.0, and the fact that two of the three demand models already do this and `RPKElasticity` does not. |
> | 9 | `compare_dict_of_arrays` handles NaN since 6.1.0 | **checked, not applicable.** That function is used for cache comparison, not for the residual, so it does not remove the need for NaN handling in the converter. The sentinel is gone for other reasons. |
> | 10 | multi-regional coverage; benchmark vs pre-#145 | **coverage done** (`test_multi_regional_process.py`, 12 tests). **Benchmark deferred** — with a note on why CI cannot see it: the timing job compares against `main`, which already contains #145. The baseline is `1f27878e^1` = `226da654`. |
> 
> Two things found while doing the above that were not in your list:
> 
> - **§2 of this PR was only half-applied.** `3bdbfb8c` tightened `_top_level_mda_chain` but
>   left `_setup_unified_mda`'s `mda_chain` at `tolerance=1e-5` with no `max_mda_iter`,
>   under a new comment claiming the settings matched. The mode with *more* coupling was
>   still on GEMSEO's 20-iteration default. Fixed, and the test that would have caught it
>   now exists.
> - **`config_elasticity_demand.yaml` is not bit-reproducible across invocations.** Two
>   fresh interpreters give final residuals of 6.192993501921497e-11 and
>   6.193005219628041e-11. Deterministic within a run; the drift is unrelated to any change
>   here (both arms of every A/B move together). It means "bit-identical" cannot be
>   established by comparing separate runs — including some comparisons earlier in this PR,
>   which should be redone in-process. Worth its own issue.
> 
> ---
> 
> ## 1. Why the airfare NaN here and not in WCTR
> 
> Two different questions with two different answers.
> 
> **The structural NaN is in WCTR too, identically.** Every elasticity scenario carries the
> same 19 NaN in `airfare_per_rpk` — 2000–2018, whatever the horizon — because
> `PassengerAircraftMarginalCost` writes 2019 explicitly and nothing writes before it:
> 19 of 51 on tutorial 08, 19 of 71 on each WCTR config; 43 of 245 and 28 of 149 resolved
> couplings carry a NaN respectively.
> 
> **What WCTR does not get is the excursion.** The spike's failure was the airfare going
> *negative*, which under a fractional exponent gives NaN silently. I tried to provoke it:
> 
> | | iterations | residual | airfare handed to the model |
> |---|---|---|---|
> | stock | 28 | 6.15e-11 | [0.0924, 0.1388] |
> | `Alternate2Delta` | 11 | 8.32e-11 | [0.0924, 0.1388] |
> | `Secant` | 16 | 4.56e-11 | [0.0924, 0.1388] |
> | `Aitken` | 13 | 3.62e-11 | [0.0924, 0.1388] |
> | `over_relaxation_factor = 1.9` | 10 | 1.45e+00 (diverging) | [0.0924, 0.1388] |
> 
> The band is [0.00924, 1.847]. Nothing gets within a factor of 13 of it, even while
> diverging. WCTR is not immune — it is benign. The excursion needs the cross-regional
> fuel-market coupling from the spike.
> 
> ---
> 
> ## 2. The sentinel is gone, and `set_bounds` replaces the in-model clip
> 
> ### It was never a performance problem
> 
> Worth saying first, because it was my assumption too and it is wrong. Profiled whole
> `compute()` runs:
> 
> | | tutorial 08 | wctr price-2019 |
> |---|---|---|
> | total | 15.65 s | 95.29 s |
> | `_iterate_once` (running the disciplines) | **67.4 %** | **83.5 %** |
> | all conversion | 3.5 % | 1.2 % |
> | dead components in the resolved vector | 12.4 % | 15.0 % |
> 
> Removing every dead component would save a fraction of an already small few percent. This
> is a correctness change.
> 
> ### Why the sentinel is bad
> 
> It puts a **missing-value flag inside the numeric channel**. `-999999` is not metadata the
> solver respects, it is a number, and the solver is entitled to do arithmetic to that
> vector. Every operation it may legitimately perform is a way to destroy the flag, and none
> of them can be told not to:
> 
> - **Blending.** Damping and acceleration recombine successive iterates. Mix `-999999` with
>   a real value and you get an arbitrary number of order 1e5–1e6 that no longer maps back
>   to NaN. Safe *only* while the NaN pattern is perfectly stable — a precondition nothing
>   checks and nothing enforces.
> - **Scaling.** `N_COUPLING_VARIABLES` residual scaling divides by `sqrt(n)` over a vector
>   that is 12–15 % dead, loosening the convergence criterion by ~2.2× without anyone
>   choosing that.
> - **Projection.** See below.
> - **Aliasing.** A genuine value of exactly `-999999.0` reads back as NaN.
> - **Detectability.** Because the flag is indistinguishable from data, "this chain died
>   mid-solve" could only be *inferred* — the check guessed from the shape of the NaNs. A
>   heuristic where an invariant belongs.
> 
> ### Why sentinel + `set_bounds` was the worst combination
> 
> `set_bounds` is enforced by projecting the **whole** resolved-variable vector
> (`SequenceTransformer._project`). The projection cannot tell a sentinel from a value, so
> `-999999` is clipped up to the lower bound, stops matching `== -999999`, never converts
> back, and is handed to the disciplines as a real number.
> 
> Three things make this the worst case rather than merely another instance:
> 
> 1. **It is inverted.** The bound is a *correctness* measure. Applying it is what destroys
>    correctness. The more careful you are, the worse the outcome.
> 2. **It is silent.** Each discipline recomputes its NaN output every iteration, so the
>    output DataFrames look untouched. Only the residual moves — and before this PR nothing
>    looked at the residual.
> 3. **It fails where you need it most.** NaN-free couplings can be bounded safely. The ones
>    carrying NaN are those spanning the historical years — most of the interesting ones —
>    so it broke precisely the scenarios stiff enough to want bounding.
> 
> Measured on tutorial 08, which makes no excursion at all and had nothing to bound: 9
> iterations at 1.27e-11 became 20 at 6.88e-07, not converged. The bounds were correctly
> placed (51 finite entries, lower bound 0.00923638). The bound broke a run that converged
> without it.
> 
> ### What replaced it
> 
> Every NaN pattern turned out to be either a leading prefix or the whole series — **zero
> scattered**, on both scenarios (21/22/0 on tutorial 08, 8/20/0 on WCTR). So it is one
> boolean mask per coupling, now held **beside** the vector. Dead positions carry `0.0`;
> both sides of the residual carry the same constant, so they difference to exactly zero —
> the same contribution the sentinel made — with nothing of magnitude 1e6 in the vector.
> 
> **When to freeze it** was the only real design decision. `BaseMDA._execute` is
> `if self._pre_solve(): self._solve()`, and `MDAGaussSeidel._pre_solve` runs **one complete
> sweep of every discipline** before `_solve`, hence before any residual is taken. That is
> the first moment every coupling has been written by its real producer. Everything earlier
> is wrong, and I checked each: the first value converted is the length-1 grammar
> placeholder; the first full-length value for `airfare_per_rpk` is the `_coupling_defaults`
> seed, NaN-free where the real series is not; and a declared horizon does not fit either,
> because the prefix lengths are {19: 1 variable, 20: 14, 21: 6} against a prospection
> offset of 20 — **per variable, not per process**.
> 
> Frozen **per solve**, not per session, and held in a `ContextVar` rather than a module
> global because `separate_processes(parallel=True)` solves regions in a
> `ThreadPoolExecutor`.
> 
> | | iterations | residual | |
> |---|---|---|---|
> | sentinel, no bounds | 9 | 1.27e-11 | converged |
> | sentinel + `set_bounds` | 20 | 6.88e-07 | not converged |
> | **mask**, no bounds | 9 | 1.27e-11 | converged |
> | **mask** + `set_bounds` | **9** | **1.27e-11** | **converged** |
> 
> Residual histories identical iteration by iteration, 657/657 columns bit-identical, worst
> relative delta `0.000e+00`. Two-region tutorial unchanged in both modes. "Converged on
> NaN" is now exact rather than heuristic.
> 
> ### The in-model clip is gone
> 
> Agreed, and done. `RPKElasticity` no longer touches the airfare it is handed. It
> **declares** its physical domain (`AIRFARE_BOUNDS_RELATIVE` → `_coupling_bounds`), the
> process hands that to `MDAChain.set_bounds`, and the solver projects its own iterate.
> 
> Why the clip was the wrong place, stated as the property it broke: **the model computed
> something other than what the solver believed it asked for.** The iterate said the airfare
> was *x*; the residual was formed on *x*; the physics ran on `clip(x)`. Where those differ,
> `r(x) = 0` no longer means the model is at equilibrium at *x* — the fixed point found
> belongs to a different function, and nothing in the residual can reveal the substitution.
> Four consequences:
> 
> 1. The solver cannot account for what it cannot see. Damping, acceleration and the
>    divergence guard all reason about a function that is not the one being evaluated.
> 2. It left no trace. `airfare_per_rpk` is an input, so a clip on it appears in no output
>    variable — the only evidence was a warning, in a notebook that emits thousands.
> 3. It is discontinuous exactly where a gradient-based scenario would read a zero Jacobian.
> 4. It is ad hoc: every model needing this reimplements its own band, warning and
>    semantics, and none of them compose.
> 
> `set_bounds` has none of those properties: the projection changes the iterate itself, so
> the value a discipline receives *is* the value the residual is formed on.
> 
> **One caveat, measured rather than assumed, and you should know it before relying on
> this.** `MDAGaussSeidel._iterate_once` runs the whole sweep and projects the transformed
> iterate *afterwards*. The bound therefore governs the iterate carried **between**
> iterations — including everything the sequence transformer extrapolates, which is what
> produced the −2.11 EUR/RPK — but **not** a value a producer hands straight to a consumer
> inside one sweep. In this chain the cost models produce `airfare_per_rpk` before
> `RPKElasticity` consumes it, so that second case is the normal one. Measured on tutorial
> 08: a band of (0.099, 0.101), far tighter than the airfares the scenario produces, changed
> **nothing at all** — the model received exactly the same [0.08925, 0.11199] with and
> without it, and the solve was bit-identical.
> 
> That is a deliberate division rather than a gap. A value the solver invented is the
> solver's to fix, and it fixes it. A value a discipline genuinely computed is a modelling
> result: if the cost chain really produces a negative airfare, the honest outcome is NaN
> and a failed run — which `check_mda_convergence` now reports exactly, via the mask — not a
> saturated number that looks plausible and hides it. Net effect versus the old clip: the
> same excursions are caught, and the ones that are not caught now fail loudly instead of
> silently returning a different model's answer.
> 
> ---
> 
> ## 3. `gemseo` pinned to `>=6.2.0,<6.3.0`
> 
> Agreed on the shape; the lower bound wants to be 6.2 rather than 6.1:
> 
> | feature | version | where |
> |---|---|---|
> | `SequenceTransformer.set_bounds` | 6.1.0 | now used, per §2 |
> | `max_consecutive_unsuccessful_iterations` | 6.1.0 | the guard `check_mda_convergence` diagnoses |
> | `compare_dict_of_arrays(nan_are_equal=…)` | 6.1.0 | — |
> | **`inner_mda_settings` as a `BaseMDASettings` model** | **6.2.0** | `_setup_unified_mda` passes it |
> 
> `poetry.lock` already resolves 6.2.0, so that is what CI tests and what every measurement
> here was made on; `>=6.1.0` would claim support for a version nothing has run. The real
> bug was that the constraint had **no lower bound at all** — an install without the lock
> could resolve 6.0.x.
> 
> Only the `content-hash` in `poetry.lock` is touched: my poetry is 2.3.1 and a full
> `poetry lock` migrates `lock-version` 2.0 → 2.1, which would break CI if
> `vars.POETRY_VERSION` is a 1.x.
> 
> ---
> 
> ## 4. Pass-through — answered, not implemented
> 
> **Why 1.0 is hardcoded: it was never chosen.** The instruments are added after the supply
> function — `marginal_cost + tax − subsidy`. There is no pass-through model; 1.0 is what a
> bare addition does. And it is the **open-loop** derivative at fixed `rpk`, which is how
> the tests call the model; inside the MDA `rpk` responds to the fare and feeds back through
> `a·rpk`, so the closed-loop figure is not 1.0 for *either* category. The ledger in the code
> documents the open-loop number and should say so.
> 
> **Two of the three demand models already compute it dynamically, and always have:**
> 
> | demand model | price it feels | delay | how fuel policy reaches it |
> |---|---|---|---|
> | `RPKElasticity` (`cagr_elasticity`) | `airfare_per_rpk`, the whole fare | **none** | via the fare, at 1.0, same year (new in §8) |
> | `PriceAndIncomeElasticity` (`constant_elasticity`) | `doc_net_energy_per_rpk_mean`, energy only | τ = 1.3312 | via the energy DOC, delayed |
> | `PriceElasticityLogisticIncome` (`logistic_income`) | `doc_net_energy_per_rpk_mean`, energy only | τ = 1.2562 | via the energy DOC, delayed |
> 
> `doc_net_energy_per_rpk_mean` is `energy + carbon_tax − subsidy + energy_tax`, so the WCTR
> models have **always** seen fuel policy, with a lag. The one model that feels the whole
> fare — the only reason §8 exists — is the one with no lag. After this PR the same energy
> tax reaches demand at three different speeds depending only on which demand model the
> scenario picked.
> 
> **The ratio, computed the way you described** — the gap between actual and consumer-felt
> cost — is two lines in any model that has a delay:
> 
> ```python
> pass_through = (perceived - reference) / (actual - reference)
> ```
> 
> normalised by each model's own reference price, so the three are comparable even though
> they feel different prices. Both series already exist in every delayed model. Published as
> `<price>_pass_through` beside the existing `<price>_delayed`.
> 
> Guard, and it is the same lesson as the ASK-weighted means: when `actual == reference`
> there is no shock and the ratio is `0/0`. Mask it to the analytic limit — the lag's step
> response `g(k) = 1 − a^(k+1)`, `a = exp(−1/τ)` — not NaN, and not a fake 1.0:
> 
> | | τ = 1.3312 | τ = 1.2562 |
> |---|---|---|
> | year 1 | 52.8 % | 54.9 % |
> | year 2 | 77.7 % | 79.7 % |
> | year 3 | 89.5 % | 90.8 % |
> | year 4 | 95.0 % | 95.9 % |
> 
> **Adding the delay to `RPKElasticity`** is three lines. Measured on tutorial 08 (prototype,
> not committed): τ = 0 bit-identical, τ = 1.3312 → +0.605 % rpk 2050, τ = 3 → +1.82 %,
> τ = 8 → +5.01 %, in 8–9 iterations throughout — the lag reduces the loop gain, so the MDA
> does not suffer.
> 
> **Two inconsistencies to settle first**, both exposed by the diagnostic: the two existing
> `_apply_price_delay` implementations are duplicated verbatim with different calibrated τ
> and both **seed on the first projected year**, so that year passes at 100 % and the lag
> only bites from year 2; and both propagate NaN through the recursion, which in an MDA
> means absorbing a transient into a stateful recursion. Both are fixed in one move by
> lifting a single `first_order_lag` into `aeromaps/models/base.py`. Own PR — it changes
> results for the two WCTR models, and τ for the fare is a calibration question.
> 
> ---
> 
> ## What is in the diff
> 
> - `pyproject.toml` / `poetry.lock` — the pin.
> - `aeromaps/core/gemseo.py` — mask registry, freeze hook, converter, `apply_coupling_bounds`,
>   exact "converged on NaN".
> - `aeromaps/core/process.py`, `multi_regional_process.py` — install both; the §2 fix.
> - `aeromaps/models/.../rpk_market.py` — clip removed, domain declared.
> - `aeromaps/models/.../ask_market.py`, `direct_operating_costs.py` — the ASK shares.
> - Tests: `test_nan_mask.py`, `test_mda_residual_floor.py`, `test_multi_regional_process.py`
>   new; defect-reproducing arms added to the airfare, cost-mean and subsidy modules.
> - `aeromaps/notebooks/dev/mda_failure_modes.py` — the reproduction script.
> 
> Not in the diff, deliberately: the price delay, the pass-through diagnostic, the logging
> clean-up, the JAX/array port, the #145 benchmark.

antSGS (commented):
> A few general comments for each problem raised:
> 
> 1. Good point the convergence, some of my WCTR applications had convergence issues that were also silent, I think that overall we can say that we need a more effective way of logging things (too many small warnings currently, and too few warnings over important stuff)
> 2. A bit difficult to understand this issue without some code to reproduce it, could you add maybe one file that demonstrates each failure mode? (even if they'll be deleted later)
> 3. Big picture stuff there, as all models can read and mutate the io dataframe this was bound to happen eventually. In my opinion, it makes sense to deal with this issue either at the same time or before we do porting to JAX (i think we'll need to get rid of dfs for computing mdas/opts and just fill the df after all was iterated in numpy/jaxarrays)
> 4. Fixed on gemseo's v6.1.0 `SequenceTransformer class now has a set_bounds method` https://gemseo.readthedocs.io/en/stable/software/changelog.html#version-6-1-0-2025-03-17
> 5. Claude already has some elements of answer there, maybe instead of computing aggregated DOC/ASK with (volumeA *qA + volumeB * qB)/(volA + volB) which yields the 0/0 you could simply make the calculation based on shares (shareA * qA + shareB * qB) even if volA+volB is zero the shares are still well defined because that's the base of computations in both fleet and energy
> 6. OK with Claude's solution, would have to be rethinked depending on what is done on 3
> 7. not much to say here
> 8. Good catch, I wonder how the pass-through is going to be modeled though... in the WCTR applications a delay is applied between airline cost and consumer-felt cost that could be used to compute a dynamic pass-through instead of a static one
> 9. Also fixed on gemseo's v6.1.0 `compare_dict_of_arrays function now handles NaNs`
> 10. (Verification): the coverage gap in the multi-region might be good to address if we are to capitalize them on the app. One thing I noticed in my notebooks is that since the generic markets PR #145 the process is now slower to compute (cumbersome especially for MDAs), might be good to re-make the benchmark from a_r_m and compare with what we had before this PR 

Thanks for the suggestions. 

3/ I agree on thinking that for later maybe. This PR should remain rather small IMO.

4/ was not used becaude of the converter and -999999 hack bit if it's okay to remove the converter because gemseo now support nan in couplings then fine. 

8/ most of claude comment is not relevant. The fact that subsidies and non-carbon fuel taxes did not reached the airfare is true. However, the assymetric behaviour between actual airlines cost and taxes is meant that way, under the hypothesis that passengers are paying the atx as a supplement (it does not affect the supply curve). More questionnable talking about the subsidy. 

10/ OKay, noted. Rtaher strange as no coupling involved in _a_r_m ... But i remember having to use a high tolerance (1e-5) when running arm back then, i can't remember why though. 


Thanks for the following commits, i'll check that now.

PR #158, opened 2026-08-20
Fix kerosene selectivity being ignored, and inverted, in the bottom-up model

antSGS:
## Summary

Kerosene selectivity was both **ignored** and **inverted** in the bottom-up environmental
model. The two defects masked each other, so the resource-budget outputs looked plausible
while being wrong for every bottom-up pathway. Found from the web app while wiring up the
MACC chart.

**Bug A — the model read a name no configuration writes.** It looked selectivity up as
`{pathway}_eis_kerosene_selectivity`. No YAML writes that: every config, bottom-up ones
included, writes plain `kerosene_selectivity` → `{pathway}_kerosene_selectivity`, the name
the top-down model already reads. The `.get()` missed and selectivity fell back to `1.0`.

**Bug B — the direction was wrong.** Bottom-up *multiplied*; top-down has always *divided*.
Division is what the quantity means, and what the docs state: selectivity is the share of a
plant's output that is aviation fuel, so 1 MJ of jet fuel at 15 % selectivity mobilises
`1/0.15 ≈ 6.67×` the feedstock, not `0.15×`. Bug A hid this, since `×1` and `÷1` agree —
fixing only the lookup would have shipped `0.15×`, understating the resource footprint by
~44×, the dangerous direction for a sustainability budget.

**The fix:** one file, nine lines. Read the same name as the top-down model (selectivity is
a plant property no config varies by commissioning year, so it needs no `eis_` prefix, and
one name for both approaches removes this class of mismatch), divide instead of multiply,
keep the existing `1.0` fallback. No configuration files change.

## Verification

Reproduction from the bug report, on the bottom-up tutorial data
(`hefa_fog` declares `kerosene_selectivity: 0.15`):

| `mobilised / consumption` | 2035 | 2045 | 2050 |
|---|---|---|---|
| before | 1.0 | 1.0 | 1.0 |
| after | 6.6667 | 6.6667 | 6.6667 |

`aeromaps/tests/models/test_kerosene_selectivity.py` pins the ratio, top-down/bottom-up
agreement, the 1.0 fallback in both models, and — end-to-end through real config plumbing —
that the key a configuration writes is the key the model reads. That last one is what
catches Bug A, which lived in the name plumbing rather than the arithmetic. Each defect was
re-introduced separately to confirm the tests fail without the fix. Full `pytest aeromaps`
passes.

## Blast radius

Resource-budget series only (`*_necessary_with_selectivity`,
`*_necessary_global_share_with_selectivity`). Publication outputs are regenerated in a
separate commit.

The committed baselines were already stale, so to isolate this change the three affected
notebooks were run twice — on `main`'s unfixed model and on this branch — and the two
*regenerated* sets diffed. **Delta attributable to the fix: 136 series across ten files,
all selectivity/resource-budget, zero others.** No CO₂, cost, traffic, fleet or climate
movement. Both `alternate_low` scenarios come back identical, a useful negative control:
their only resource-consuming pathways run at selectivity 1.0, where `×1` and `÷1` agree.

**`joas_2023` would otherwise have failed CI** — its notebook compares against
`data/reference/outputs.json` and raises beyond `rtol=1e-4`, and it runs in the watchdog
workflow. Reference regenerated, notebook re-run to confirm it passes.

**On the pre-existing staleness in the diff:** the committed publication outputs had drifted
from `main` — `_2019` → `_last_historical_year`, `corsia_reference_year`, `freight_covid_*`,
various growth rates — because earlier merges landed model changes without re-executing and
re-committing the notebook outputs. Regenerating sweeps that up, so part of this diff is
that backlog rather than this fix. `joas_2023`'s reference had stayed current, which is what
made the clean attribution above possible. `aeromaps_data.xlsx` is a `write_excel()` output
of the same run.

## Also included

`pyproject.toml` `1.0.0` → `1.1.0`, missed when 1.1.0 was cut. Separate commit, drop it if
you want this single-purpose. Note the released wheel was always fine —
`poetry-dynamic-versioning` builds from the tag, so PyPI has `aeromaps-1.1.0`; only source
and editable installs reported `1.0.0`, which is presumably what the web app runs.

Noted, not fixed: `docs/books/CHANGELOG.md` is missing the 1.1.0 section that the root
`CHANGELOG.md` has.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


iancostalves:
> Bug A — the model read a name no configuration writes.

Hm interesting ... I had a similar problem with this sort of behaviour and was iterating it on my atag branch. But instead, I think it makes sense to commit it here instead, I'll document the issue here, and can work on it after incorporating the review on atag.

iancostalves:
# Misspelled or misplaced YAML input keys silently resolve to zero

## Summary

Every input the generic energy models read is looked up with `input_data.get(key, default)`, where
`default` is a zero-filled series. A key that does not match — because it is misspelled, because it
lost a prefix, or because it sits in a block the reading model does not register — therefore
resolves to **zero** rather than raising.

Zero is a physically meaningful value for an emission factor, a cost or a consumption rate. Nothing
downstream complains, no warning is emitted, and the run completes with plausible-looking numbers.

We have now hit this **four times in production data and found three more latent instances in the
code**. The largest of them understated a scenario's 2050 CO₂ by roughly 40 % for an unknown length
of time. This issue proposes making the failure loud.

## Evidence

### Instances that produced wrong published numbers

| # | Location | Key written | Key read | Effect |
|---|---|---|---|---|
| 1 | all 12 ATAG `*_energy.yaml` | `co2_emission_factor_without_resource` | `..._mean_co2_emission_factor_without_resource` | All seven biomass SAF pathways modelled as **zero-carbon**. Scenario S1 2050 CO₂ read 386 Mt; correct value 644 Mt. |
| 2 | `top_down/environmental.py:287` | (data was correct) | process lookup missing the `mean_` prefix | Process emissions dropped **repo-wide**. Hydrogen liquefaction and electrolysis both read exactly 0.0000. |
| 3 | ATAG `*_energy.yaml` | intensity curves anchored later than the mandate | — | SAF produced at **zero cost and zero emissions** in the gap years. Five years wide in the 2nd edition. |
| 4 | `mea_2024/energy_carriers_data.yaml:388` | `resources_names` | `resource_names` | `hydrogen_electrolysis`'s `transport` resource silently dropped from cost and emissions. |

\#1 and \#2 are fixed. \#3 is fixed (see "Related fix" below). **\#4 is not fixed** — it is still
present on `main`. That file's own schema comment at line 25 uses the singular, as do its other
seven declarations; line 388 is the odd one out.

### Latent instances already written into the code

None of these fire today, because no current scenario uses subsidies, taxes, or the bottom-up
models. Each is the same mechanism, waiting.

| Location | Code reads | Loader/template produces |
|---|---|---|
| `top_down/cost.py:287` | `{process}_mean_unit_subsidy_without_resource`**s** | `processes_data.yaml:26` → `..._mean_unit_subsidie`**s**`_without_resource` |
| `top_down/cost.py:295` | `{process}_mean_unit_tax_without_resource`**s** | `processes_data.yaml:30` → `..._without_resource` |
| `top_down/cost.py:225,259` | `{resource}_subsidy` | `resources_data.yaml:23` documents `subsidies` |
| `bottom_up/environmental.py:405` | `{process}_eis_co2_emission_factor_without_resource`**s** | template and the top-down twin both use `..._without_resource` |

### A related gap that is live for any new user

`common/energy_use_choice.py:65-76` dispatches on the mandate type with
`if "quantity" ... elif "share" ...` and **no `else: raise`**. The shipped template documents

```yaml
mandate_type: "share"    # or "volume"
```

but the code never tests for `"volume"`. Anyone following the documented template gets a pathway
with no mandate at all, silently — it simply never deploys. `mandate_quantity`, the value the code
actually accepts, is undocumented.

## Why the obvious fix is wrong

**Making the lookups strict does not work.** Many of these inputs are genuinely optional, and this
was measured rather than assumed: of **505 pathways** across the repository, only **310** declare
`mean_co2_emission_factor_without_resource`. The other **195** legitimately omit it —
`fossil_kerosene` and the bottom-up carriers derive their emission factor elsewhere. Requiring the
key would break all of them.

## Proposed fix: reject unknown keys at load time

The distinguishing signal is that **a misspelling produces an _unknown_ key, whereas a legitimately
absent optional input produces _no_ key at all.** Those two cases are separable at load time, which
a strict `.get()` at lookup time cannot do.

In `aeromaps/utils/yaml.py`, when parsing an energy-carriers, processes or resources file:

1. Validate each block's keys (`mandate`, `technical`, `environmental`, `economics`) against an
   allow-list.
2. Raise on any key outside it, naming **file, carrier, block and key**.
3. Attach a **"did you mean"** suggestion via `difflib.get_close_matches`. Every instance above is a
   near-miss — a dropped `mean_` prefix, a stray plural — so this turns each one into a one-line
   actionable error.

### Build the allow-list from the code, not from the templates

The commented templates in `aeromaps/resources/data/default_energy_carriers/` are the closest thing
to a schema the project has, and they are **demonstrably not authoritative**: they disagree with the
code in the four places tabulated above, plus the `mandate_type` value. Deriving the allow-list from
them would encode the bugs it is meant to catch.

The observed vocabulary is small and closed:

- **environmental** — `mean_co2_emission_factor_without_resource`,
  `eis_co2_emission_factor_without_resource`, `emission_index`
- **technical** — `lhv`, `kerosene_selectivity`, `resource_names`, `processes_names`,
  `resource_specific_consumption`, `plant_lifespan`, `plant_load_factor`,
  `technology_introduction_year`, `technology_introduction_quantity` / `_volume`, and the `eis_` twins
- **economics** — `mean_mfsp_without_resource`, `eis_capex`, `eis_fixed_opex`, `eis_variable_opex`,
  `mean_unit_subsidy_without_resource`, `mean_unit_tax_without_resource`

### Block placement: a narrower rule than it first appears

Verified mechanics, which matter for getting this right:

- `_flatten_dict` (`utils/functions.py:64-72`) **discards the block name** — the prefix is the
  pathway name. So `environmental:` and `technical:` yield identical flattened keys.
- The environmental model registers `environmental` **+ `technical`**; the cost model registers
  `economics` **+ `technical`** (`top_down/environmental.py:53-64`, `top_down/cost.py:60-71`).

Consequently:

- A key under `technical:` is visible to **both** models. The `ecats_2026` files that put
  `mean_co2_emission_factor_without_resource` there are **harmless — do not "fix" them.**
- An emission-factor key under `economics:`, or an MFSP key under `environmental:`, is registered by
  only the *other* model and silently reads zero. **That** is the case worth rejecting.

### Also in scope

Add the missing `else: raise` to the `mandate_type` dispatch, and correct the template to document
`"quantity"` and `mandate_quantity`.

## Blast radius

**Expect the validator to fail on existing files** — instance \#4 at minimum, and possibly others
not yet found. That is the point, but it means the rollout order is:

1. Add the validator behind a flag, or as a warning.
2. Sweep the whole repository and fix every file it flags.
3. Re-run every publication and diff committed outputs; any movement is a bug this found.
4. Turn the validator into a hard error.

Publications to re-check: `wctr_2026`, `ecats_2026`, `icas_2024`, `joas_2023`, `mea_2024`,
`salgas_phd_2025`, and the ATAG scenario tree.

## Verification

1. Each of the four historical typos is rejected with a "did you mean" naming the correct key — one
   unit test per shape (dropped prefix, stray plural, wrong block).
2. A pathway legitimately omitting `mean_co2_emission_factor_without_resource` still loads. There
   are 195 such pathways; none may start failing.
3. Whole-repo load sweep: every `*energy*.yaml`, `processes*.yaml` and `resources*.yaml` parses
   clean once the offenders are fixed.
4. `mandate_type: "volume"` raises instead of silently producing a mandate-less pathway.
5. Full suite green.

antSGS (commented):
> # Misspelled or misplaced YAML input keys silently resolve to zero
> 
> ## Summary
> 
> Every input the generic energy models read is looked up with `input_data.get(key, default)`, where
> `default` is a zero-filled series. A key that does not match — because it is misspelled, because it
> lost a prefix, or because it sits in a block the reading model does not register — therefore
> resolves to **zero** rather than raising.
> 
> Zero is a physically meaningful value for an emission factor, a cost or a consumption rate. Nothing
> downstream complains, no warning is emitted, and the run completes with plausible-looking numbers.
> 
> We have now hit this **four times in production data and found three more latent instances in the
> code**. The largest of them understated a scenario's 2050 CO₂ by roughly 40 % for an unknown length
> of time. This issue proposes making the failure loud.
> 
> ## Evidence
> 
> ### Instances that produced wrong published numbers
> 
> | # | Location | Key written | Key read | Effect |
> |---|---|---|---|---|
> | 1 | all 12 ATAG `*_energy.yaml` | `co2_emission_factor_without_resource` | `..._mean_co2_emission_factor_without_resource` | All seven biomass SAF pathways modelled as **zero-carbon**. Scenario S1 2050 CO₂ read 386 Mt; correct value 644 Mt. |
> | 2 | `top_down/environmental.py:287` | (data was correct) | process lookup missing the `mean_` prefix | Process emissions dropped **repo-wide**. Hydrogen liquefaction and electrolysis both read exactly 0.0000. |
> | 3 | ATAG `*_energy.yaml` | intensity curves anchored later than the mandate | — | SAF produced at **zero cost and zero emissions** in the gap years. Five years wide in the 2nd edition. |
> | 4 | `mea_2024/energy_carriers_data.yaml:388` | `resources_names` | `resource_names` | `hydrogen_electrolysis`'s `transport` resource silently dropped from cost and emissions. |
> 
> \#1 and \#2 are fixed. \#3 is fixed (see "Related fix" below). **\#4 is not fixed** — it is still
> present on `main`. That file's own schema comment at line 25 uses the singular, as do its other
> seven declarations; line 388 is the odd one out.
> 
> ### Latent instances already written into the code
> 
> None of these fire today, because no current scenario uses subsidies, taxes, or the bottom-up
> models. Each is the same mechanism, waiting.
> 
> | Location | Code reads | Loader/template produces |
> |---|---|---|
> | `top_down/cost.py:287` | `{process}_mean_unit_subsidy_without_resource`**s** | `processes_data.yaml:26` → `..._mean_unit_subsidie`**s**`_without_resource` |
> | `top_down/cost.py:295` | `{process}_mean_unit_tax_without_resource`**s** | `processes_data.yaml:30` → `..._without_resource` |
> | `top_down/cost.py:225,259` | `{resource}_subsidy` | `resources_data.yaml:23` documents `subsidies` |
> | `bottom_up/environmental.py:405` | `{process}_eis_co2_emission_factor_without_resource`**s** | template and the top-down twin both use `..._without_resource` |
> 
> ### A related gap that is live for any new user
> 
> `common/energy_use_choice.py:65-76` dispatches on the mandate type with
> `if "quantity" ... elif "share" ...` and **no `else: raise`**. The shipped template documents
> 
> ```yaml
> mandate_type: "share"    # or "volume"
> ```
> 
> but the code never tests for `"volume"`. Anyone following the documented template gets a pathway
> with no mandate at all, silently — it simply never deploys. `mandate_quantity`, the value the code
> actually accepts, is undocumented.
> 
> ## Why the obvious fix is wrong
> 
> **Making the lookups strict does not work.** Many of these inputs are genuinely optional, and this
> was measured rather than assumed: of **505 pathways** across the repository, only **310** declare
> `mean_co2_emission_factor_without_resource`. The other **195** legitimately omit it —
> `fossil_kerosene` and the bottom-up carriers derive their emission factor elsewhere. Requiring the
> key would break all of them.
> 
> ## Proposed fix: reject unknown keys at load time
> 
> The distinguishing signal is that **a misspelling produces an _unknown_ key, whereas a legitimately
> absent optional input produces _no_ key at all.** Those two cases are separable at load time, which
> a strict `.get()` at lookup time cannot do.
> 
> In `aeromaps/utils/yaml.py`, when parsing an energy-carriers, processes or resources file:
> 
> 1. Validate each block's keys (`mandate`, `technical`, `environmental`, `economics`) against an
>    allow-list.
> 2. Raise on any key outside it, naming **file, carrier, block and key**.
> 3. Attach a **"did you mean"** suggestion via `difflib.get_close_matches`. Every instance above is a
>    near-miss — a dropped `mean_` prefix, a stray plural — so this turns each one into a one-line
>    actionable error.
> 
> ### Build the allow-list from the code, not from the templates
> 
> The commented templates in `aeromaps/resources/data/default_energy_carriers/` are the closest thing
> to a schema the project has, and they are **demonstrably not authoritative**: they disagree with the
> code in the four places tabulated above, plus the `mandate_type` value. Deriving the allow-list from
> them would encode the bugs it is meant to catch.
> 
> The observed vocabulary is small and closed:
> 
> - **environmental** — `mean_co2_emission_factor_without_resource`,
>   `eis_co2_emission_factor_without_resource`, `emission_index`
> - **technical** — `lhv`, `kerosene_selectivity`, `resource_names`, `processes_names`,
>   `resource_specific_consumption`, `plant_lifespan`, `plant_load_factor`,
>   `technology_introduction_year`, `technology_introduction_quantity` / `_volume`, and the `eis_` twins
> - **economics** — `mean_mfsp_without_resource`, `eis_capex`, `eis_fixed_opex`, `eis_variable_opex`,
>   `mean_unit_subsidy_without_resource`, `mean_unit_tax_without_resource`
> 
> ### Block placement: a narrower rule than it first appears
> 
> Verified mechanics, which matter for getting this right:
> 
> - `_flatten_dict` (`utils/functions.py:64-72`) **discards the block name** — the prefix is the
>   pathway name. So `environmental:` and `technical:` yield identical flattened keys.
> - The environmental model registers `environmental` **+ `technical`**; the cost model registers
>   `economics` **+ `technical`** (`top_down/environmental.py:53-64`, `top_down/cost.py:60-71`).
> 
> Consequently:
> 
> - A key under `technical:` is visible to **both** models. The `ecats_2026` files that put
>   `mean_co2_emission_factor_without_resource` there are **harmless — do not "fix" them.**
> - An emission-factor key under `economics:`, or an MFSP key under `environmental:`, is registered by
>   only the *other* model and silently reads zero. **That** is the case worth rejecting.
> 
> ### Also in scope
> 
> Add the missing `else: raise` to the `mandate_type` dispatch, and correct the template to document
> `"quantity"` and `mandate_quantity`.
> 
> ## Blast radius
> 
> **Expect the validator to fail on existing files** — instance \#4 at minimum, and possibly others
> not yet found. That is the point, but it means the rollout order is:
> 
> 1. Add the validator behind a flag, or as a warning.
> 2. Sweep the whole repository and fix every file it flags.
> 3. Re-run every publication and diff committed outputs; any movement is a bug this found.
> 4. Turn the validator into a hard error.
> 
> Publications to re-check: `wctr_2026`, `ecats_2026`, `icas_2024`, `joas_2023`, `mea_2024`,
> `salgas_phd_2025`, and the ATAG scenario tree.
> 
> ## Verification
> 
> 1. Each of the four historical typos is rejected with a "did you mean" naming the correct key — one
>    unit test per shape (dropped prefix, stray plural, wrong block).
> 2. A pathway legitimately omitting `mean_co2_emission_factor_without_resource` still loads. There
>    are 195 such pathways; none may start failing.
> 3. Whole-repo load sweep: every `*energy*.yaml`, `processes*.yaml` and `resources*.yaml` parses
>    clean once the offenders are fixed.
> 4. `mandate_type: "volume"` raises instead of silently producing a mandate-less pathway.
> 5. Full suite green.
>  

Thanks for the good, related issues. As you point out, most of the difficulty lies in teh maximal genericity asked for this project. Thus some keys are genuinely optional. 
Claude fix seems rather good, maybe the "Attach a "did you mean" suggestion" is a bit too much though ? 

Should i tkae care of making teh suggested edits or you want to take care of it @iancostalves?

antSGS:
## Implemented: rejecting unknown energy-YAML keys (`a218f90f`)

> *This comment, and the commit it describes, were written by Claude (Claude Code), driven by @antSGS. Please review accordingly.*

This addresses @iancostalves' issue above. Two deviations from the proposal, both deliberate — see "Departures" at the end.

### What lands

A validator, `generic_energy_model/common/yaml_schema.py`, called from the three places `process.py` reads energy files. Hard error, not a warning (see the sweep below for why the staged rollout wasn't needed).

**The allow-list is derived from the models, not transcribed.** Every energy model now declares two class attributes:

```python
class TopDownEnvironmental(AeroMAPSModel):
    PATHWAY_INPUT_BLOCKS = ("environmental", "technical")
    PATHWAY_INPUT_KEYS = ("mean_co2_emission_factor_without_resource", "emission_index", ...)
```

The schema **unions the keys** to decide what is *known*, and **intersects the registered blocks of every model that reads a key** to decide *where it may sit*. A key must be visible to every one of its readers, and a model only ever sees the blocks it registers — so the intersection *derives* the block rule instead of restating it:

| key | readers | allowed blocks |
|---|---|---|
| `mean_co2_emission_factor_without_resource` | environmental models only | `environmental`, `technical` |
| `mean_mfsp_without_resource` | cost models only | `economics`, `technical` |
| `eis_plant_lifespan` | BU environmental + cost + capacity | `technical` only |
| `mandate_type` | `EnergyUseChoice` | `mandate` only |

This gives the narrow rule you identified for free: the `icas_2024` pathways that put an emission factor under `technical:` stay valid (both models register that block), while the same key under `economics:` is rejected. Adding an input to a model now means adding it to that model's own declaration, so the list cannot drift from the code.

Nested mappings are checked too: `emission_index` species, `availability` sub-keys, and — same failure mode — a `resource_specific_consumption` entry naming a resource the pathway doesn't declare.

### The sweep

**117 referenced energy files, one offender** — exactly instance #4, and both halves of it:

```
mea_2024/data/energy_carriers_data.yaml:
  - hydrogen_electrolysis.inputs.technical.resources_names: unknown key, no energy model reads it.
  - hydrogen_electrolysis.inputs.technical.resource_specific_consumption.transport:
    'transport' is not one of the resources 'hydrogen_electrolysis' declares in 'resource_names',
    so this consumption is never read.
```

Fixed. Blast radius measured by running that config both ways: **24 series, all in that pathway's transport-cost chain** — hydrogen MFSP up to +5.7 %, energy expenses +1.9 %, six series that previously didn't exist, **no CO₂ movement** (the `transport` resource declares a cost, no emission factor). Notebook re-runs clean; it's stored stripped with no committed `outputs_*.json`, so there's nothing to regenerate.

Because the sweep found exactly one file rather than an unknown set, steps 1–3 of your staged rollout collapse into this commit. **No other publication needs re-running.**

### Three read names corrected first

A list derived from the models would have rejected *every* `processes_data.yaml` in the repo, because the models read names no config writes:

| location | read | every config writes |
|---|---|---|
| `top_down/environmental.py` | `{process}_co2_emission_factor_without_resource`, and the process `environmental:` block was **never registered** | `mean_co2_emission_factor_without_resource` |
| `bottom_up/environmental.py` | `..._eis_co2_emission_factor_without_resource`**s**, same unregistered block | `eis_co2_emission_factor_without_resource` |
| `top_down/cost.py` | `{process}_mean_unit_subsidy`/`tax_without_resource`**s** | nothing yet (latent) |

The first two are your instance #2. **Every process emission factor in the repository is `0.0` and no configuration declares process subsidies, so no committed number moves** — full suite confirms it. ⚠️ @iancostalves, this duplicates the #2 fix already on `atag_scenario`; that branch will conflict on those lines and should take this side.

Also in scope, as you asked: `else: raise` on the `mandate_type` dispatch, and the templates corrected to document `"quantity"` / `mandate_quantity` and to use `mean_unit_subsidy_without_resource` and the resource `subsidy` key the code actually reads.

### Verification

19 tests in `test_energy_yaml_schema.py`, each historical shape reproduced next to the case it must not break: dropped `mean_` prefix, stray plural, wrong block (both directions), misspelled species, undeclared resource — against optional inputs legitimately absent, empty blocks, both mandate types, and the `technical:` placement that must stay valid. Plus the whole-repo sweep and one end-to-end test that corrupts a file and asserts `create_process` refuses it.

Each fix was re-broken separately to confirm the matching test fails without it. `pytest aeromaps`: **193 passed**.

### Departures from the proposal

1. **No "did you mean" via `difflib`** — dropped at @antSGS' request. The error names file, entry, block and key, and lists the keys accepted in that block, which covers the same ground deterministically.
2. **Straight to a hard error**, no warning phase — justified by the sweep finding a single offender.

Happy to split the three read-name corrections into their own commit if you'd rather review them separately from the validator.

iancostalves (approved):
OK by me, may just need a re-push of the updated json outputs for the notebook validation

antSGS:
**Correction to my comment above** (`a6363ff2`) — *written by Claude, driven by @antSGS.*

Block placement is now **exact**: each key belongs to one block, declared by the model that reads it. My earlier claim that a key under `technical:` "stays valid" no longer holds — that flexibility was dropped at @antSGS' request.

Consequently the three `icas_2024` files move `fossil_kerosene`'s emission factor from `technical:` to `environmental:`. Running that configuration both ways leaves **every output series identical**, since the model that reads it registers both blocks.

One follow-on from the process-emission fix: the factor now resolves to the interpolated series, which starts at the prospection start year (the process files declare it with an empty `years:`), and the addition into the pathway emission factor had no `fill_value` — so the shorter index turned the pathway's historical years into NaN. Added `fill_value=0`, same as `atag_scenario`. The per-process diagnostic series now reads NaN instead of a fabricated 0.0 before 2020, where no configuration defines a value. Everything else moves by at most 2.5e-15.

`pytest aeromaps`: 195 passed. Reference outputs are being refreshed separately by @antSGS.

